PR #7, opened 2019-06-27
Update CalSys* scripts to use salobj v4 and schema configuration

couger01:
- fixed flake8 violations

### DIFF
--- a/python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
+++ b/python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
@@ -28,7 +28,7 @@ class CalSysTakeNarrowbandData(scriptqueue.BaseScript):
                                "auxiliary telescope CalSystem.")
         self.cmd_timeout = 60
         self.change_grating_time = 60
-        self.electrometer = salobj.Remote(domain=self.domain, name="Electrometer")
+        self.electrometer = salobj.Remote(domain=self.domain, name="Electrometer", index=1)
         self.monochromator = salobj.Remote(domain=self.domain, name="ATMonochromator")
         self.fiber_spectrograph = salobj.Remote(domain=self.domain, name="FiberSpectrograph")
         self.atcamera = salobj.Remote(domain=self.domain, name="ATCamera")
@@ -145,7 +145,7 @@ class CalSysTakeNarrowbandData(scriptqueue.BaseScript):
         required: [wavelengths, integration_times, fiber_spectrograph_integration_times]
         additionalProperties: false
         """
-        return yaml_schema
+        return yaml.safe_load(yaml_schema)
 
     async def configure(self, config):
         """Configure the script.
--- a/python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
+++ b/python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 import asyncio
-import collections
 
 import numpy as np
 
@@ -9,12 +8,7 @@ from lsst.ts import salobj
 from lsst.ts import scriptqueue
 from .calsys_takedata import is_sequence, as_array
 
-import SALPY_ATMonochromator
-import SALPY_Electrometer
-import SALPY_FiberSpectrograph
-import SALPY_ATCamera
-import SALPY_ATSpectrograph
-import SALPY_ATArchiver
+from lsst.ts.idl.enums import ATMonochromator
 import csv
 import datetime
 import os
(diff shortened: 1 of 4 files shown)


python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
```diff
@@ -23,42 +17,137 @@
...
-                                       'monochromator': salobj.Remote(SALPY_ATMonochromator),
-                                       'fiber_spectrograph': salobj.Remote(SALPY_FiberSpectrograph),
-                                       'atcamera': salobj.Remote(SALPY_ATCamera),
-                                       'atspectrograph': salobj.Remote(SALPY_ATSpectrograph),
-                                       'atarchiver': salobj.Remote(SALPY_ATArchiver)
-                                      })
+                               "auxiliary telescope CalSystem.")
         self.cmd_timeout = 60
         self.change_grating_time = 60
+        self.electrometer = salobj.Remote(domain=self.domain, name="Electrometer")
```
r-owen: I think you need `index=1` for the electrometer,  based on reading the original code.

python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
```diff
@@ -23,42 +17,137 @@
...
+          file_location:
+            type: string
+            default: "~/develop"
+          script_type:
+            type: string
+            default: "narrowband"
+        required: [wavelengths, integration_times, fiber_spectrograph_integration_times]
+        additionalProperties: false
+        """
+        return yaml_schema
```
r-owen: The schema must be returned as a `dict`. See ts_standardscripts calsys_takedata.py for an example of how to do this.

r-owen: Also you are replacing a configure call that allowed scalars for values with a schema that does not.

I see plusses and minuses to this approach. On the plus side:
- The schema is significantly simpler.
- The input rules for users are simpler: always specify an array.
On the down side:
- The configuration is a bit messier for users because they have to specify arrays.
- It may feel a bit odder to have the code turn [x] into [x, x, ...x] as opposed to turning a scalar x into [x, x, ...x]

It would be a big win if all the cal scripts worked the same way, so could you please talk to @tribeiro about this. As the architect I'd prefer he sign off on it. (I am happy to join the conversation) and decide what you prefer want.

ts_standardscripts calsys_takedata.py uses the more flexible schema and would be easy to copy for this script. On the other hand if you end up decide you prefer the regularity of always specifying an array for every item then I'm happy to simplify the standard script to match.

I am happy to join the conversation if wanted.

python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
```diff
@@ -99,39 +188,42 @@ def __init__(self, index):
...
                 nelt = len(value)
                 break
-        self.file_location=os.path.expanduser(file_location)
-        self.setup_spectrograph = setup_spectrograph
-        self.take_image = take_image
-        self.script_type = script_type
+        self.file_location = os.path.expanduser(self.config.file_location)
+        self.setup_spectrograph = self.config.setup_spectrograph
+        self.take_image = self.config.take_image
+        self.script_type = self.config.script_type
```
r-owen: I very strongly recommend *not* copying data from config to separate fields in the script. Just use `self.config.script_type` when you want the script type. Copying adds a lot of boilerplate and is error-prone. Note that the schema constrains the data types in the config much as a class definition would, so it's not as if you'll be getting arbitrary nonsense data that you have to validate.

python/lsst/ts/externalscripts/auxtel/calsys_take_narrowband_data.py
```diff
@@ -311,23 +404,26 @@ def set_metadata(self, metadata):
...
-                                                 expTime=self.integration_times[index],
-                                                 imageSequenceName=self.image_sequence_name[index])
-        atarchiver_lfo_coro = self.atarchiver.evt_processingStatus.next(flush=True,timeout=(self.cmd_timeout*2)+self.integration_times[index])
+        self.atcamera.cmd_takeImages.set(
+            shutter=self.shutter[index],
+            numImages=1,
+            expTime=self.integration_times[index],
+            imageSequenceName=self.image_sequence_name[index])
+        atarchiver_lfo_coro = self.atarchiver.evt_processingStatus.next(flush=True, timeout=(
+            self.cmd_timeout*2)+self.integration_times[index])
```
r-owen: I suspect this (pre-existing) code code not do what you think it does. The flush in evt_processingStatus.next will not occur until you start awaiting it. I think what you want is as follows (I also combined the set and start, to save a line, but explicitly calling flush first is the important change):

    self.atarchiver.evt_processingStatus.flush()
    await self.atcamera.cmd_take_images.set_start(
       shutter=...
    )
    return self.atarchiver.evt_processingStatus.next(flush=True, timeout=...)

PR #13, opened 2020-02-28
Added camera_hexapod Simulator 3.1.1 test script

couger01:


tribeiro (changes requested):
@couger01 can you clarify what is the purpose of this script? Is this supposed to be a SAL Script to run on the ScriptQueue? It definitively doesn't look like one. It seems like it is more like a "bin script" to run on the command line? If that is the case, not sure it belongs here.

tribeiro:
Closing after talking to Eric in private.

PR #17, opened 2020-04-14
Tickets/dm 24309

patrickingraham:


python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -68,24 +71,23 @@ class LatissCWFSAlign(salobj.BaseScript):
...
     """
 
     __test__ = False  # stop pytest from warning that this is not a test
 
-    def __init__(self, index, remotes=True):
+    def __init__(self, index=1, remotes=False):
 
         super().__init__(
-            index=index, descr="Perform optical alignment procedure on LATISS data."
+            index=index, descr="Perform optical alignment procedure on LATISS in-focus images using Curvature-Wavefront Sensing Techniques."
```
tribeiro, line 90: Are you sure "in-focus image" is the correct term here? Maybe just

>Perform optical alignment procedure on LATISS images using Curvature-Wavefront Sensing Techniques.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -230,14 +250,18 @@ def dz(self, value):
...
             fp.write(
                 cwfs_config_template.format(self._dz * 0.041, 10e-6 * self.binning)
             )
+
         self.inst = Instrument(config_index, int(self.side * 2 / self.binning))
         self.algo = Algorithm("exp", self.inst, 1)
 
     async def take_intra_extra(self):
-        """ Take Intra/Extra images.
+
```
tribeiro, line 262: Remove space between method and docstring.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -385,46 +430,55 @@ async def run_cwfs(self):
...
         self.algo.reset(self.I1[0], self.I2[0])
 
+        # for a slow telescope, should be running in paraxial mode
         await loop.run_in_executor(
-            executor, self.algo.runIt, self.inst, self.I1[0], self.I2[0], "onAxis"
+            executor, self.algo.runIt, self.inst, self.I1[0], self.I2[0], "paraxial"
         )
-        # self.algo.runIt(self.inst, self.I1[0], self.I2[0], 'onAxis')
 
+        self.log.warning(" TODO: Negative sign being used infront of calculated Coma-X Zernike! Artifact tha appeared while checking at NCSA. Needs investigation on other platforms!")
```
tribeiro, line 463: Maybe add a comment instead of a warning?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -483,6 +537,7 @@ def sum_source_flux(_source, exp, min_size):
 
     def center_and_cut_images(self):
         """ After defining sources for cwfs cut snippet for cwfs analysis.
+        This is used to speed up the CWFS analysis code
```
tribeiro, line 540: It is not exactly to speed up the CWFS but to provide a snapshot with the donut for it to analyze.

tests/auxtel/test_latiss_cwfs_align.py
```diff
@@ -0,0 +1,42 @@
+import logging
```
tribeiro, line 1: I don't think this unit test is doing anything. Consider removing it.

patrickingraham: Doesn't this verify that the script is executable at a minimal level?
In the workflow procedure we wrote there is a minimum requirement of having a test that proves it compiles etc.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: Changed to: 
Perform optical alignment procedure of the Rubin Auxiliary Telescope with LATISS using Curvature-Wavefront Sensing Techniques

tests/auxtel/test_latiss_cwfs_align.py
tribeiro: The way it is right now it is not testing that there is an executable. Also, I don't think we have an executable yet and the run method on the class is still empty. So, even if we had and executable, it would not do much. To see how to test that the executable exists take a look at this one:

https://github.com/lsst-ts/ts_standardscripts/blob/cf710a07381fcf854e7aea7236ba558a79cf6f1a/tests/auxtel/test_startup.py#L50-L53

In any case, I think the best now would be to remove the test and let it the way it is right now. We should give it some thought and think about how to implement the run method and then proceed to create the executable and the test.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: Not sure.. this is a pretty serious issue...

tests/auxtel/test_latiss_cwfs_align.py
patrickingraham: Ok done, but lets try to get something in place this week as a standard. With the operations rehearsal coming up we'll want them following the proper protocol.

patrickingraham:
incorporated all comments, except the removal of the warning. Also rebased and squashed.

PR #21, opened 2020-09-23
Tickets/dm 26495

patrickingraham:
PR is more to assist discussion at this point.

### DIFF
--- a/python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
+++ b/python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
@@ -20,33 +20,38 @@
 
 __all__ = ["LatissAcquireAndTakeSequence"]
 
-import yaml
-import warnings
 import asyncio
-import numpy as np
+import collections.abc
+import warnings
 
+import lsst.daf.persistence as dafPersist
+import numpy as np
+import yaml
+import concurrent.futures
+from astropy import time as astropytime
+from lsst.geom import PointD
 from lsst.ts import salobj
 from lsst.ts.observatory.control.auxtel import ATCS, LATISS
+from lsst.ts.observatory.control.constants import latiss_constants
+from lsst.ts.observatory.control.utils import RotType
+
+from lsst.ts.standardscripts.utils import format_as_list
 
 try:
-    from lsst.observing.utils.audio import playSound
-    from lsst.observing.utils.filters import changeFilterAndGrating, getFilterAndGrating
-    from lsst.observing.utils.offsets import findOffsetsAndMove
-    from lsst.observing.constants import boreSight, sweetSpots
-except ImportError:
-    warnings.warn("Cannot import lsst.observing. Script will not work.")
-
-
-try:
-    import lsst.daf.persistence as dafPersist
-except ImportError:
-    warnings.warn(
-        "Cannot import 'lsst.daf.persistence.dafPersist'. Script will not work."
+    from lsst.ts.observing.utilities.auxtel.latiss.utils import (
+        calculate_xy_offsets,
+        parse_obs_id,
     )
+    from lsst.pipe.tasks.quickFrameMeasurement import QuickFrameMeasurementTask
+    from lsst.ts.observing.utilities.auxtel.latiss.getters import get_image
+except ImportError:
+    warnings.warn("Cannot import required libraries. Script will not work.")
+
+STD_TIMEOUT = 10  # seconds
 
 
 class LatissAcquireAndTakeSequence(salobj.BaseScript):
-    """ TODO: update docs
+    """
     Perform an acquisition of a target on LATISS with the AuxTel.
     This sets up the instrument and puts the brightest target on a
     specific pixel, then takes a sequence of exposures for a given
@@ -59,7 +64,8 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
 
     Notes
     -----
-    **Checkpoints**
+    An (optional) checkpoint is available to verify the calculated
+    telescope offset after each iteration of the acquisition.
 
     **Details**
 
@@ -73,11 +79,21 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
     def __init__(self, index, silent=False):
 
         super().__init__(
-            index=index, descr="Perform target acquisition for LATISS instrument."
+            index=index,
+            descr="Perform target acquisition and data taking"
+            " for LATISS instrument.",
         )
 
-        self.attcs = ATCS(self.domain)
-        self.latiss = LATISS(self.domain)
+        self.atcs = ATCS(self.domain, log=self.log)
+        self.latiss = LATISS(self.domain, log=self.log)
+        # instantiate the quick measurement class
+        try:
+            qm_config = QuickFrameMeasurementTask.ConfigClass()
+            self.qm = QuickFrameMeasurementTask(config=qm_config)
+        except NameError:
+            self.log.warning("Library unavailable certain tests will be skipped")
+        # Set timeout
+        self.cmd_timeout = 30  # [s]
 
         # Suppress verbosity
         self.silent = silent
@@ -105,6 +121,13 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                 description: SIMBAD query-able object name
                 type: string
 
+              manual_focus_offset:
+                description: Applies manual focus offset after the slew. This
+                    is temporary in order to observe low-altitude stars until
+                    the ATAOS LUTs improve.
+                type: number
+                default: 0.0
+
               acq_filter:
                 description: Which filter to use when performing acquisition. Must use
                              filter = BG40 for now
@@ -123,14 +146,22 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                 default: 2.
 
               max_acq_iter:
-                description: Max number of iterations to perform when putting source in place.
+                description: Max number of iterations to perform when acquiring target at a location.
+                             Only used if do_acquire=True.
                 type: number
                 default: 3
+                minimum: 1
 
               target_pointing_tolerance:
                 description: Number of arcsec from source to desired position to consider good enough.
                 type: number
                 default: 5
+
+              target_pointing_verification:
+                description: Take a follow-up exposure to verify calculated offset was applied
+                    correctly before starting sequence?
+                type: boolean
+                default: True
 
               filter_sequence:
                 description: Filters for exposure sequence. If a single value is specified then
@@ -155,8 +186,8 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                 default: empty_1
 
               exposure_time_sequence:
-                description: Exposure times for exposure sequence (sec). If a single value
-                  is specified then the same exposure time is used for each exposure.
+                description: Exposure times for exposure sequence (sec). Each exposure requires
+                   a specified exposure time.
                 anyOf:
                   - type: array
                     minItems: 1
@@ -173,7 +204,7 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                 default: /project/shared/auxTel/
 
               doPointingModel:
-                description: Adjust star position (sweet-spot) to use boresight
+                description: Adjust star position (sweet spot) to use boresight
                 type: boolean
                 default: False
 
@@ -209,6 +240,8 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
         self.acq_grating = config.acq_grating
         self.acq_filter = config.acq_filter
         self.acq_exposure_time = config.acq_exposure_time
+        self.manual_focus_offset = config.manual_focus_offset
+        self.manual_focus_offset_applied = False
 
         # Max number of iterations to perform when putting source in place
         self.max_acq_iter = config.max_acq_iter
@@ -216,6 +249,7 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
         # Tolerance in arcsec for distance from main source centroid to
         # relevant sweet spot
         self.target_pointing_tolerance = config.target_pointing_tolerance
+        self.target_pointing_verification = config.target_pointing_verification
 
         self.acq_visit_config = (
             self.acq_filter,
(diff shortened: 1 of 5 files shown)


python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -177,6 +188,41 @@ def get_schema(cls):
         """
         return yaml.safe_load(schema_yaml)
 
+    def format_as_array(self, value, recurrences):
```
tribeiro, line 191: This seems like a general purpose utility method. Consider moving it to https://github.com/lsst-ts/ts_standardscripts/blob/develop/python/lsst/ts/standardscripts/utils.py

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -219,18 +266,23 @@ async def configure(self, config):
         )
 
         # make a list of tuples from the filter, exptime and grating lists
+        if isinstance(config.exposure_time_sequence, collections.Iterable):
```
tribeiro, line 269: you can make this in a single line like:

```
_recurrences = len(config.exposure_time_sequence) if isinstance(config.exposure_time_sequence, collections.Iterable) else 1
```

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -240,116 +292,188 @@ def set_metadata(self, metadata):
             gratings.add(grating)
         metadata.filter = f"{filters},{gratings}"
 
-    async def latiss_acquire(self, silent=True, doPointingModel=False):
+    async def latiss_acquire(self, doPointingModel=False):
+
+        # instantiate the quick measurement class
+        qm = QuickFrameMeasurement()
```
tribeiro, line 298: consider making this a class attribute:

```
self.qm = QuickFrameMeasurement()
```

in the `__init__` method.

patrickingraham: So i'll do this, in fact I had it that way originally, except I thought that since it was only required in one of the functions that instantiating it for everything (when it might not get used) was not the right thing to do.
Is the "best practice" to put this in the `__init__` even if it won't be used?

patrickingraham:
I think all changes have been incorporated. Tests are failing due to the `rapid_analysis` package not being setup, which I imagine is not part of the container... so I'm not sure how to get around that... it's over my head. I suppose I could catch the exception and skip all tests, but is that what we want to do?

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -240,116 +256,191 @@ def set_metadata(self, metadata):
...
-                    doMove=True,
-                    alwaysAcceptMove=_alwaysAcceptMove,
-                ),
+        # If ATAOS is running wait for adjustments to complete before
+        # moving on.
+        if corr.atspectrograph:
+            self.log.debug(f"Verifying LATISS configuration is incorporated into ATAOS offsets")
+            # If so, then flush correction events for confirmation of
+            # corrections
+            await self.atcs.rem.ataos.evt_atspectrographCorrectionStarted.aget(timeout=self.cmd_timeout)
```
tribeiro, line 293: Are you sure you want to use `aget` here and in the following line? you are flushing the events above so I think you may want to use `next` with `flush=False`...

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -240,116 +256,191 @@ def set_metadata(self, metadata):
...
+            "Entering Acquisition Iterative Loop, with a maximum amount of "
+            f"iterations set to {self.max_acq_iter}"
+        )
+        iter_num = 0
+        while iter_num < self.max_acq_iter:
+            # Take image
+            self.log.debug(f"Starting iteration number {iter_num+1}")
+            tmp, data_id = await asyncio.gather(
                 self.latiss.take_object(exptime=self.acq_exposure_time, n=1),
+                get_next_image_dataId(self.latiss, timeout=self.acq_exposure_time + STD_TIMEOUT,),
```
tribeiro, line 306: So, `take_object` returns the dataId, but seems like you want to wait for the `atarchiver.evt_imageInOODS` event. I think this could be incorporated to the latiss class. It could be an option on `take_object` for instance.

patrickingraham: So we could add it to `take_object`, but we will start slewing as soon as the image it done, but we can't do anything with the image until it's in the OODS.... so I think we need to keep these functionalities separate?
It's true that maybe `get_next_image_dataId` might not be capturing exactly what the method is doing (which is really saying when you can pull the image).

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -240,116 +256,191 @@ def set_metadata(self, metadata):
...
-            dr_arcsec = np.sqrt(dx_arcsec ** 2.0 + dy_arcsec ** 2.0)
-            print(f"Calculated offsets [dx,dy] are [{dx_arcsec},{dy_arcsec}] arcsec")
-            print(f"Current radial pointing error is {dr_arcsec}")
+            # Find brightest star
+            result = self.qm.run(exp)
+            current_position = PointD(result.brightestObjCentroid[0], result.brightestObjCentroid[1])
+
+            # Find offsets
+            self.log.debug(f"Current brightest target position is {current_position}")
+            self.log.debug(f"Target position is {target_position}")
```
tribeiro, line 322: Consider merging both debug calls into one.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
tribeiro: uhm... not sure I understood your comment. What you mean by "we will start slewing as soon as the image is done"? Isn't this software performing an acquisition where you take an image, analyze it and perform offsets?

If you are talking in more general terms, my proposal is to add a new parameter, e.g. `wait_ingestion` that, if `True` would wait for the OODS event and return the data id. 

By default `wait_ingestion` would be `False` and the method retain the current behavior.

tribeiro: So, it may be worth dealing with this particular feature in the future. Maybe just move `get_next_image_dataId` to be a method of this class and we implement it later in ts_observatory_control.

patrickingraham: ok.
You're correct about this software needing to wait for image analysis, but other use-cases certainly won't wait, which is why incorporating it into take_object doesn't address all use-cases...

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: uhh what's the difference?
Is it that if multiple events are received then `next` picks the one right after the flush whereas `aget` takes the most recent?

tribeiro: Yes. So, if you get here before the new event is published you get the last one seen (with `aget`), even with the flush above. Whereas `next` will give you what ever was published after the flush.

patrickingraham: right, because `aget` and `get` aren't affected by flushes... good catch.

patrickingraham:
Changes made. Also fixed some DM variable naming. 
Note that the `rapid_analysis` dependency issue remains unresolved.
Please "resolve Conversation" where applicable, just so it's easier to read what is outstanding.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -73,11 +79,21 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
...
-        self.attcs = ATCS(self.domain)
-        self.latiss = LATISS(self.domain)
+        self.atcs = ATCS(self.domain, log=self.log)
+        self.latiss = LATISS(self.domain, log=self.log)
+        # instantiate the quick measurement class
+        try:
+            qm_config = QuickFrameMeasurementTask.ConfigClass()
+            self.qm = QuickFrameMeasurementTask(config=qm_config)
+        except NameError:
+            self.log.warning("Library unavailable ertain tests will be skipped")
```
tribeiro, line 94: ertain -> certain ?

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -245,116 +285,371 @@ def set_metadata(self, metadata):
...
+        # await asyncio.gather(
+        #     self.atcs.slew_object(name=self.object_name, rot=0,
+        #     slew_timeout=240),
+        #     self.latiss.setup_instrument(grating=self.acq_grating,
+        #     filter=self.acq_filter),
+        # )
+
+        # Slew and setup in series for now
+        # set rot_type to align to paralactic angle
+        await self.atcs.slew_object(name=self.object_name, rot_type=2, slew_timeout=240)
```
tribeiro, line 345: Consider importing the enumeration and using it here:

At top do:
```
from lsst.ts.observatory.control.utils import RotType
```

Then here do:

```
await self.atcs.slew_object(name=self.object_name, rot_type=RotType.Parallactic., slew_timeout=240)

```

The instrument setup is done later (in line 70), but the code you comment above is doing it in parallel with `asyncio.gather`... May be a good idea to either do them in sequence here:

```
await self.atcs.slew_object(name=self.object_name, rot_type=2, slew_timeout=240)
await self.latiss.setup_instrument(grating=self.acq_grating, filter=self.acq_filter),
```

or leave a comment here so you remember to remove the code in line 370 once persistent offset it working.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -245,116 +285,371 @@ def set_metadata(self, metadata):
...
+                )
 
-        if success:
-            self.log.info("Achieved centering accuracy")
-        else:
-            if (
-                iterNum != -1
-            ):  # also heinous to support max_acq_iter = 0 working TODO: fix this!
+            # Ask user if we want to apply the correction?
+            await self.checkpoint(
```
tribeiro, line 474: I think this is the only checkpoint in the script... might be a good idea to add a couple more....

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -245,116 +285,371 @@ def set_metadata(self, metadata):
...
+                f"{self.manual_focus_offset} in after acquisition"
             )
-            if not silent:
-                playSound("ding" if i < nexp - 1 else "gong")
+            await self.atcs.rem.ataos.cmd_offset.set_start(z=-self.manual_focus_offset)
+            self.manual_focus_offset_applied = False
 
     async def run(self):
         """"""
         if self.do_acquire:
```
tribeiro, line 633: Consider adding a couple checkoutpoint in the run method.

tests/auxtel/test_latiss_acquire_and_take_sequence.py
```diff
@@ -0,0 +1,414 @@
...
+
+from lsst.ts import salobj
+from lsst.ts import standardscripts
+from lsst.ts import externalscripts
+from lsst.ts.externalscripts.auxtel import LatissAcquireAndTakeSequence
+import lsst.daf.persistence as dafPersist
+
+random.seed(47)  # for set_random_lsst_dds_domain
+
+logging.basicConfig(level=logging.DEBUG)
```
tribeiro, line 37: Do you really need to do this logging stuff? you can get the logging info from the unit test running 

```
pytest -vs --log-level DEBUG
```

tribeiro (changes requested):
Made a couple comments in line. Overall looks good, just take a quick look on the comments and we are ready to merge.

tests/auxtel/test_latiss_acquire_and_take_sequence.py
patrickingraham: i got rid of the log level issue, but the matplotlib stuff is somewhat required or it's impossible to diagnose anything

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: Done... but i'll revisit this in the next ticket

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: i'm confused... the gather statement there is commented out because of the persistent offset issue...
I added a comment, does that address the issue?

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -245,116 +285,380 @@ def set_metadata(self, metadata):
...
+            # automatically...
+
             self.log.debug("Beginning taking data for target sequence")
+            await self.checkpoint("Beginning taking data for target sequence")
             try:
                 await self.latiss_take_sequence()
+                self.checkpoint("Data taking for target sequence completed")
             except Exception as e:
-                print("grabbed exception from latiss_take_sequence(): ")
+                self.log.warning("grabbed exception from latiss_take_sequence(): ")
```
tribeiro, line 661: Consider using:

```
self.log.exception("Exception from latiss_take_sequence()")
```

So that the traceback is logged.

tribeiro (approved):
Looks good! Made a small comment to use a `self.log.exception` instead of `self.log.warning`.... feel free to merge...

PR #27, opened 2021-03-12
Tickets/dm 29061 - Updates to latissAcquireAndTakeSequence. Also deprecation of two scripts that have been incorporated into this one.

patrickingraham:


python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -464,8 +474,8 @@ async def latiss_acquire(self):
                         z=-self.manual_focus_offset
                     )
                     self.manual_focus_offset_applied = False
-                self.log.exception(e)
-                raise e
+                self.log.exception("Centroid finding algorithm was unsuccessful.")
```
tribeiro, line 477: I think you want to use `self.log.error` here... exception is for when you are logging an exception that was raised before.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -464,8 +474,8 @@ async def latiss_acquire(self):
...
+                raise RuntimeError("Centroid finding algorithm was unsuccessful.")
```
tribeiro, line 478: If you are raising the exception here, you probably don't need to the log message above.

tribeiro (approved):
See inline comments.

PR #29, opened 2021-04-14
Tickets/dm 29387

patrickingraham:


tests/auxtel/test_latiss_cwfs_align.py
```diff
@@ -0,0 +1,512 @@
...
+        nasmyth_angle = self.visit_id_angles[self.script.extra_visit_id][1]
+        bore_sight_angle = elevation_angle + parity_x * nasmyth_angle + 90.0
+
+        return bore_sight_angle
+
+    @unittest.skipIf(
+        DATA_AVAILABLE is False,
+        "Data not available for test_take_sequence, skipping.",
+    )
+    async def test_take_sequence(self):
```
patrickingraham, line 204: We really want to run this on several datasets.. not just one.

tests/auxtel/test_latiss_cwfs_align.py
```diff
@@ -0,0 +1,512 @@
...
+                    <= 0.05
+                )
+                or (abs(self.script.total_focus_offset - total_focus) < hex_tol[2])
+            )
+
+    @unittest.skipIf(
+        DATA_AVAILABLE is False,
+        "Data not available for test_analysis, skipping.",
+    )
+    async def test_analysis(self):
```
patrickingraham, line 387: same here I think.
You'll see that this runs on images 307 and 308, where the one above runs on 307, 308 and 309,310 but only verifies the latter and the totals.

patrickingraham:
Tests will fail as it takes a branch of pipe-tasks to complete successfully.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -432,33 +374,42 @@ async def run_cwfs(self):
...
 
-        self.extra_exposure = await loop.run_in_executor(
-            executor, self.get_isr_exposure, self.extra_visit_id
+        self.intra_result = await loop.run_in_executor(
+            executor,
+            functools.partial(
+                self.qm.run, self.intra_exposure, donutDiameter=2 * self.side
+            ),
         )
+        self.extra_result = await loop.run_in_executor(
```
tribeiro, line 391: What does the `self.qm.run` do? is it doing isr plus source detection? If it is doing the source selection, why do it on both? Wouldn't it be best to do the source selection on an image that is intra + extra, the way we did before?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -432,33 +374,42 @@ async def run_cwfs(self):
...
-        self.intra_exposure = await loop.run_in_executor(
-            executor, self.get_isr_exposure, self.intra_visit_id
+        self.intra_exposure, self.extra_exposure = await asyncio.gather(
+            get_image(parse_visit_id(self.intra_visit_id)),
+            get_image(parse_visit_id(self.extra_visit_id)),
         )
 
-        # create a clone to be used for detection
-        self.detection_exp = self.intra_exposure.clone()
+        # Run source detection - FIXME: make this run concurrently
```
tribeiro, line 382: You are already logging "Run source detection" I think you can leave just the "FIXME" part.

tests/auxtel/test_latiss_cwfs_align.py
tribeiro: Although I agree it really makes me wonder, how much of this is testing the script and how much is testing the source selection algorithm. I think we have to separate these two things. I don't think we should expect Scripts to re-test external libraries.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: It does just the source finding. The ISR is handled in `get_image`
The reason I switched away from adding the two together is mostly for simplicity. The source selection (if run in parallel) is pretty quick and doesn't add time. 
There is also the effect that simply adding the doughnuts together (which are shifted a bit) makes a weird shape... although I don't think this really matters.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -338,7 +320,7 @@ async def hexapod_offset(self, offset, x=0.0, y=0.0):
 
         Returns
         -------
-
+        None
```
tribeiro, line 323: If return is `None`, then you can remove this section from the docstring.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -359,51 +341,11 @@ async def hexapod_offset(self, offset, x=0.0, y=0.0):
...
         )
 
-    def get_isr_exposure(self, exp_id):
-        """Get ISR corrected exposure.
+    async def run_cwfs(self):
+        """Runs CWFS code on intra/extra focal images.
 
         Parameters
         ----------
-        exp_id: `string`
```
tribeiro, line 367: If there are no input parameters, remove the section for the docstring.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -467,172 +418,117 @@ async def run_cwfs(self):
...
-        """
+    def create_donut_stamps_for_cwfs(self):
+        """Create square stamps with donuts based on centroids.
 
-        if self.detection_exp is None:
-            raise RuntimeError(
-                "No detection exposure defined. Run take_intra_extra or"
-                "manually define sources before running."
-            )
+        Parameters
```
tribeiro, line 433: If there are no input parameters, remove sections from docstring.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -467,172 +418,117 @@ async def run_cwfs(self):
...
+        _rebinned_array = arr.reshape(shape).mean(-1).mean(1)
+        return _rebinned_array
 
     def calculate_results(self, silent=False):
         """Calculates hexapod and telescope offsets based on
         derotated zernikes.
 
-        Inputs:
-        -------
+        Paramters
```
tribeiro, line 523: If there are no parameter, remove section.

tribeiro (changes requested):
Thanks for the updates. Overall I like the direction this is going and it is exposing some issues we have in writing these scripts. It is clear to me that Scripts are the affluent where all our software will meet and cause major turmoils, so bear with me.

Here is my initial round of comments.

Regarding the changes in the `latiss_cwfs_align` script itself I have some concerns over the changes that where implemented. For instance, The source detection is now done on each individual image instead of on a combined "detection" image. I think this will make the script slower, as it has to run the source detection twice on images with lower SNR. The other issue I see is that you could potentially end up with brighter sources that are different in one image than the other. Say, for instance, if there are two stars with similar brightness in the FoV. 

My other main concern is with regard to the unit testing. As I mentioned inline how much of this is testing the Script and how much is testing the library it uses? Are we using the library in such a non-standard way that we have to perform some unit tests of our own? I think the solution would be to move those tests to the library itself. 

I think this also shows how urgent it is for us to be able to replace cwfs by wep pipeline task.

tests/auxtel/test_latiss_cwfs_align.py
patrickingraham: Discussed in person, will create a new repo for these types of analyses in the future. This script will remain unchanged since it'll be significantly refactored once the MTAOS is ready.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: Discussed in person. Will perform two source finding algorithms as there is a concern about if the algorithm works on combined donuts (which may be odd shapes).
Will add a check to make sure the results are within ~100 pixels of each other.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: Actually removed the FIXME because we decided running concurrently was not worth it.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -432,33 +374,42 @@ async def run_cwfs(self):
...
-        # Run source detection
-        await loop.run_in_executor(executor, self.select_cwfs_source)
-
-        await loop.run_in_executor(executor, self.center_and_cut_images)
-
-        if self.binning != 1:
-            self.log.info(
-                f"Running CWFS code using images binned by {self.binning} in each dimension."
+        # Verify a result was achieved, if not then raising the exception
+        if not self.intra_result.success:
```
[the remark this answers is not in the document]
patrickingraham: addressed above.
I did merge it into one...

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -467,172 +418,117 @@ async def run_cwfs(self):
...
-        def sum_source_flux(_source, exp, min_size):
-            bbox = _source.getFootprint().getBBox()
-            if bbox.getDimensions().x < min_size or bbox.getDimensions().y < min_size:
-                return -1.0
-            return np.sum(exp[bbox].image.array)
-
-        selected_source = 0
-        flux_selected = sum_source_flux(
-            result.sources[selected_source], self.detection_exp, min_size=self.pre_side
+        ceny, cenx = int(self.extra_result.brightestObjCentroidCofM[1]), int(
```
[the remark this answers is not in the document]
patrickingraham: That shouldn't be possible over a small FoV.
All donuts have the same aberrations.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -642,10 +538,12 @@ def calculate_results(self, silent=False):
 
         self.log.info(
             f"""==============================
-Measured [coma-X, coma-Y, focus] zernike coefficients in nm : {self.zern}
-De-rotated zernike coefficients: {rot_zern}
-Hexapod offset : {hexapod_offset}
-Telescope offsets: {tel_offset}
+Measured [coma-X, coma-Y, focus] zernike coefficients [nm]: [{
+            (len(self.zern) * '{:0.1f}, ').format(*self.zern)}]
```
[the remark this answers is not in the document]
patrickingraham: I can't actually... the values returned are a list, not an array.
I stole this off stack overflow about how to format a list of values (but then massaged it to fit on one line).
https://stackoverflow.com/questions/2762058/format-all-elements-of-a-list/2762087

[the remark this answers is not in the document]
patrickingraham: If it's an issue I could make a function...

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -467,172 +418,117 @@ async def run_cwfs(self):
...
             new_shape[0],
             arr.shape[0] // new_shape[0],
             new_shape[1],
             arr.shape[1] // new_shape[1],
         )
-        return arr.reshape(shape).mean(-1).mean(1)
+        _rebinned_array = arr.reshape(shape).mean(-1).mean(1)
+        return _rebinned_array
 
     def calculate_results(self, silent=False):
```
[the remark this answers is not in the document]
tribeiro: Are you going to remove the parameter as well?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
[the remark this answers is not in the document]
tribeiro: I think you can break it down into two steps. First you create the strings and then pass them to the f-string. Something like:

```
log_zern = (len(self.zern) * "{:0.1f} ").format(*self.zern)
log_rot_zern = (len(self.rot_zern) * "{:0.1f} ").format(*self.rot_zern)
log_hex_offset = (len(hexapod_offset) * "{:0.1f} ").format(*hexapod_offset)
log_tel_offset = (len(tel_offset) * "{:0.1f} ").format(*tel_offset)

self.log.info(
            f"""==============================
Measured [coma-X, coma-Y, focus] zernike coefficients [nm]: [{log_zern}]
De-rotated [coma-X, coma-Y, focus]  zernike coefficients [nm]: [{log_rot_zern}]
Hexapod [x, y, z] offsets [mm] : {(log_hex_offset}
Telescope offsets [arcsec]: {log_tel_offset}
==============================
"""
        )
```

tribeiro (approved):
I think there's a couple of comments that still needs addressing, but nothing major. Please take a look before merging.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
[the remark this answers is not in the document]
patrickingraham: Sure, but how does this help simplify anything? it only creates variables that then get printed in a log message... they're not used elsewhere?
I'll do it if you really want, I'm just not really seeing what it adds.

PR #31, opened 2021-06-07
updated tests in cwfs to use the ready_to_take_image functionality.

patrickingraham:
Added blind offset to the script in the slew command.
Removed asynchonous image taking and atarchiver event listening as it was causing failues.
Added flush before taking image to avoid race condition
Updated unit tests and code to have the ATAOS automatic offsets be handled by the latiss class
Removed rotator restrictions and now it is set to track the parallactic angle


tribeiro (approved):
looks good

PR #32, opened 2021-06-11
Tickets/dm 30638

tribeiro:


python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -751,11 +762,23 @@ async def arun(self, checkpoint=False):
...
                 if self.offset_telescope:
-                    tel_x, tel_y = results["tel_offset"][0], -results["tel_offset"][1]
-                    self.log.info(f"Applying telescope offset x/y: {tel_x}/{tel_y}.")
-                    await self.atcs.offset_xy(
-                        x=tel_x,
-                        y=tel_y,
+                    tel_el_offset, tel_az_offset = (
+                        results["tel_offset"][0],
+                        -results["tel_offset"][1]
+                        * np.cos(
```
patrickingraham, line 768: How are we supposed to know why you're doing this?
Maybe add a comment saying that the cosine is accounting for how az offsets depend on elevation.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -114,16 +114,27 @@ def get_schema(cls):
...
                 type: boolean
-                default: True
+                default: False
 
               object_name:
                 description: SIMBAD query-able object name
                 type: string
+                default: ""
+                
+              object_ra:
```
tribeiro, line 129: To allow this to be None do:

```
              object_ra:
                description: RA as a string (in hours).
                default: null
                anyOf:
                  - type: number
                  - type: null
```


same for dec bellow.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -114,16 +114,27 @@ def get_schema(cls):
...
               do_take_sequence:
                 description: Take sequence of data on target?
                 type: boolean
-                default: True
+                default: False
 
               object_name:
                 description: SIMBAD query-able object name
                 type: string
+                default: ""
```
tribeiro, line 127: I think you can require users to provide an object_name. To do that, remove the default field.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -248,8 +259,10 @@ async def configure(self, config):
         self.do_blind_offset = config.do_blind_offset
 
         # Object name
-        assert config.object_name is not None
+        assert (config.object_name is not None or (config.object_ra is not None and config.object_dec))
```
tribeiro, line 262: There is a missing None in object_dec

```
assert (config.object_name is not None or (config.object_ra is not None and config.object_dec is not None))
```

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -352,14 +365,33 @@ async def latiss_acquire(self):
         # Setup instrument and telescope
         # The following code sets up the instrument and slews the telescope
         # simultaneously.
-        tmp, data = await asyncio.gather(
-            self.atcs.slew_object(
+        if self.object_name:
```
tribeiro, line 368: I think you want should check that ra and dec are not None, if so, then use object name, e.g.;

```
if self.object_ra is not None and self.object_dec is not None:
    # call slew_icrs, but pass in the name of the target as `target_name` so it 
    # goes to the header.

    ....
else:
    # Slew object
```

tests/auxtel/test_latiss_acquire_and_take_sequence.py
```diff
@@ -345,7 +345,9 @@ async def test_take_acquisition_pointing(self):
             # sequence number start
             self.seq_num_start = 188
 
-            object_name = "HD145600"
+            # object_name = "HD145600"
```
tribeiro, line 348: uncomment `object_name`

tests/auxtel/test_latiss_acquire_and_take_sequence.py
```diff
@@ -354,7 +356,8 @@ async def test_take_acquisition_pointing(self):
             do_take_sequence = False
             do_pointing_model = True
             await self.configure_script(
-                object_name=object_name,
+                object_ra=object_ra,
```
tribeiro, line 359: Pass object name here as well.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: re-arranged this section a bit. Essentially tries to use coords and falls back to object name.
Else statement removed since object_name is now mandatory.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -142,9 +142,10 @@ def __init__(self, index=1, remotes=False):
 
         # Matrix to map hexapod offset to alt/az offset in the focal plane
         # units are arcsec/mm. X-axis is Elevation
+        # Measured with data from AT run SUMMIT-5027, still unverified.
         self.hexapod_offset_scale = [
-            [60.0, 0.0, 0.0],
-            [0.0, 60.0, 0.0],
+            [56.8, 0.0, 0.0],
```
tribeiro, line 147: Can you update these with the values that Craig provided?

 52.46

and 

50.47

respectively

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: object name not required

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: cosine removed. this comment no longer applies

PR #33, opened 2021-06-25
DM-30483: Generate calibrations at NCSA/summit with ComCam data/simulators via OCPS

plazas:


python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        """Set estimated duration of the script.
+        """
+        # Temporary number
+        metadata.duration = 10
+
+    async def run(self):
+        # Should ComCam and OCPS not be enabled here in the script?
+        # original notebook says: "When writing a script, the components
+        # should (probably) be enabled by a user."
+        await self.comcam.enable()
```
tribeiro, line 111: I would strongly suggest to not to this here in the script... If you want to, you can add a verification step that will check it comcam is enabled and fail the script if it is not (same for OCPS).

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        $schema: http://json-schema.org/draft-07/schema#
+        $id: https://github.com/lsst-ts/ts_externalscripts/blob/master/python/lsst/ts/externalscripts/>-
+        comcam/make_comcam_bias.py
+        title: MakeComCamBias v1
+        description: Configuration for making a LSSTComCam master bias SAL Script.
+        type: object
+        additionalProperties: false
+        required: [instrument, inputCollections]
+        properties:
+            nBias:
```
tribeiro, line 59: I would suggest you use `lower_snake_case` for configuration parameters.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+                type: tuple
+                items:
+                    type: integer
+                    minItems: 1
+                default: (0)
+                descriptor: Detector IDs
+
+            instrument:
+                type: string
+                enum: ["LSSTcComCam", "LATISS", "LSSTCam"]
```
tribeiro, line 74: typo?

`LSSTcComCam` -> `LSSTComCam`

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+                descriptor: Camara name. Must be LSSTComca for now (future: also LATISS or LSSTCam).
```
tribeiro, line 75: typo?

`LSSTComca` -> `LSSTComcam`.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+
+            detectors:
+                type: tuple
+                items:
+                    type: integer
+                    minItems: 1
+                default: (0)
+                descriptor: Detector IDs
+
+            instrument:
```
tribeiro, line 72: Is this really an option for this script? Its name is "make_comcam_bias" so I assume the instrument is ComCam only no?

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+
+    async def run(self):
+        # Should ComCam and OCPS not be enabled here in the script?
+        # original notebook says: "When writing a script, the components
+        # should (probably) be enabled by a user."
+        await self.comcam.enable()
+        await salobj.set_summary_state(self.ocps, salobj.State.ENABLED,
+                                       settingsToApply="LSSTComCam.yaml")
+
+        # Take config.nBiases biases, and return a list of IDs
```
tribeiro, line 115: you may want to add a `checkpoint` before taking the bias.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        tempBiasList = await self.comcam.take_bias(self.config.nBias)
```
tribeiro, line 116: we adopt `snake_case` for variables. I'd suggest `temp_bias_list`. But you can probably get rid of this and only use `exposures`...

```
exposures = tuple(await self.comcam.take_bias(self.config.nBias))
```

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        exposures = tuple(tempBiasList)
+        # Bias IDs
+        await self.checkpoint(f"Biases taken: {tempBiasList}")
+
+        # did the images get archived and are they available to the butler?
+        val = await self.comcam.rem.ccarchiver.evt_imageInOODS.aget(timeout=10)
```
tribeiro, line 122: If I understood correctly you want to wait for all the images to get ingested. I imagine that if you make sure the "last" image was ingested you can assume "all" images where ingested.  In this case, you need to verify that the image you got here is really the last one.. and if it is not, you would need to wait for another event until you got the one you need.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        await self.checkpoint(f"Biases in ccarchiver: {val}")
+
+        # Now run the bias pipetask via the OCPS
+        ack = await self.ocps.cmd_execute.set_start(
```
tribeiro, line 126: it doesn't looks like you are doing anything interesting while you wait for the command to execute. I'd just run the command with `wait_done=True` (default) and not bother with this.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        tempBiasList = await self.comcam.take_bias(self.config.nBias)
+        exposures = tuple(tempBiasList)
+        # Bias IDs
+        await self.checkpoint(f"Biases taken: {tempBiasList}")
+
+        # did the images get archived and are they available to the butler?
+        val = await self.comcam.rem.ccarchiver.evt_imageInOODS.aget(timeout=10)
+        await self.checkpoint(f"Biases in ccarchiver: {val}")
+
+        # Now run the bias pipetask via the OCPS
```
tribeiro, line 125: You probably want to flush `self.ocps.evt_job_result` before you run this command or you have several things in the event queue that you don't need to get and check afterwards. Just do:

```
self.ocps.evt_job_result.flush()
```

here

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+        ack = await self.ocps.cmd_execute.next_ackcmd(ack)
+        self.log.debug('Received command completion acknowledgement from ocps')
+        if ack.ack != salobj.SalRetCode.CMD_COMPLETE:
+            ack.print_vars()
+        # Wait for the job result message that matches the job id we're
+        # interested in ignoring any others (from other remotes).
+        # This obviously needs to follow the first acknowledgement
+        # (that returns the, job id) but might as well wait for the second.
+        while True:
+            msg = await self.ocps.evt_job_result.next(flush=False)
```
tribeiro, line 152: Doesn't seems like you are doing anything with this message afterwards. Do you really need to wait for this event from the OCPS? If so, you may want to add a timeout here or this can block indefinitely.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
```diff
@@ -0,0 +1,157 @@
...
+            response = json.loads(msg.result)
+            if response["jobId"] == job_id:
+                break
+
+        await self.checkpoint(f"Final status: {response}")
```
tribeiro, line 157: I don't think a checkpoint here does any good. If you really want to log the result consider using;

```
self.log.info(f"Final status: {response}")
```

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
plazas: Yes, that sounds good. How can I add that verification step?

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
plazas: Sounds good. I'm used to using camelCase from the DM coding guidelines.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
plazas: That's right; I'll delete it. It could be brought back in the future if/when generalize the code.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
plazas: I'll check if the retrieved ingested ID corresponds to the last image ID; if not, I'll wait for a couple more events more. If it still doesn't coincide, I'll log a warining.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
plazas: I added a timeout=600.

python/lsst/ts/externalscripts/comcam/make_comcam_bias.py
plazas: Adding `wait_done=True` actually broke things, so I'll just revert back to how it was.

python/lsst/ts/externalscripts/base_make_bias.py
```diff
@@ -0,0 +1,104 @@
...
+
+        self.config = config
+
+    def set_metadata(self, metadata):
+        """Set estimated duration of the script.
+        """
+        # Temporary number
+        metadata.duration = 10
+
+    async def arun(self, checkpoint=False):
```
tribeiro, line 99: You probably want to decorate this method with 

```
@abc.abstractmethod
```

But if you think carefully about it, I am pretty sure you can develop a single `arun` method

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,163 @@
...
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["MakeComCamBias"]
+
+import json
+import yaml
+import os
+
+from ..base_make_bias import BaseMakeBias
```
tribeiro, line 27: Can you import this last?

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,163 @@
...
+        super().__init__(
+            index=index,
+            descr="This class takes biases with LSSTComCam and constructs "
+                  "a master bias calling the bias pipetask via OCPS.",
+        )
+        self.comcam = ComCam(domain=self.domain, log=self.log)
+        self.ocps = salobj.Remote(domain=self.domain, name="OCPS")
+
+    @classmethod
+    def get_schema(cls):
```
tribeiro, line 49: I don't think you need to override this method from the parent class. If you really want to override it because of the tule and description, make sure you do:

```
        schema_dict = yaml.safe_load(schema)

        base_schema_dict = super(TakeImageComCam, cls).get_schema()

        for prop in base_schema_dict["properties"]:
            schema_dict["properties"][prop] = base_schema_dict["properties"][prop]

        return schema_dict
```

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,163 @@
...
+        schema = """
+        $schema: http://json-schema.org/draft-07/schema#
+        $id: https://github.com/lsst-ts/ts_externalscripts/blob/master/python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
+        title: MakeComCamBias v1
+        description: Configuration for making a LSSTComCam master bias SAL Script.
+        type: object
+        """
+        return yaml.safe_load(schema)
+
+    async def configure(self, config):
```
tribeiro, line 59: You don't need to override the parent class method.

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,163 @@
...
+        # Log information about the configuration
+
+        self.log.debug(
+            f"n_bias: {config.n_bias}, detectors: {config.detectors}, "
+            f"instrument: LSSTComCam. "
+        )
+
+        self.config = config
+
+    def set_metadata(self, metadata):
```
tribeiro, line 77: You don't need to override the parent class method.

python/lsst/ts/externalscripts/base_make_bias.py
```diff
@@ -0,0 +1,104 @@
...
+            f"n_bias: {config.n_bias}, detectors: {config.detectors}, "
+        )
+
+        self.config = config
+
+    def set_metadata(self, metadata):
+        """Set estimated duration of the script.
+        """
+        # Temporary number
+        metadata.duration = 10
```
tribeiro, line 97: You can probably make a best estimate than this. I'd say something like:

```
metadata.duration = config.n_bias * self.camera.read_out_time + self.estimated_process_time
```

and add `self.estimated_process_time`  to `__init__` with an estimated process time. If it is image dependent you could do:

```
metadata.duration = config.n_bias * (self.camera.read_out_time + self.estimated_process_time)
```

python/lsst/ts/externalscripts/base_make_bias.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Parameters
+    ----------
+    index : `int`
+        SAL index of this script
+    """
+
+    def __init__(self, index, descr):
+        super().__init__(index=index, descr=descr)
+
```
tribeiro, line 40: You probably want to create an abstract property that returns the "camera", e.g.;

```
    @property
    @abc.abstractmethod
    def camera(self):
        raise NotImplementedError()
```

Then, on the child classes you would do:

```
def __init__(self, index, descr):
    ...
    self._comcam = ComCam(...)
    ...

@property
def camera(self):
    return self._comcam
```

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,163 @@
...
+    a master bias calling the bias pipetask via OCPS.
+    """
+
+    def __init__(self, index=1):
+        super().__init__(
+            index=index,
+            descr="This class takes biases with LSSTComCam and constructs "
+                  "a master bias calling the bias pipetask via OCPS.",
+        )
+        self.comcam = ComCam(domain=self.domain, log=self.log)
```
tribeiro, line 45: See my comment above in the parent class. You can make an abstract property that points to this object.

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,163 @@
...
+        self.ocps = salobj.Remote(domain=self.domain, name="OCPS")
```
tribeiro, line 46: This could go to the parent class.

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self.config = config
+
+    def set_metadata(self, metadata):
+        """Set estimated duration of the script.
+        """
+        # Temporary number
+        metadata.duration = 10
+
+    async def arun(self, checkpoint=False):
```
tribeiro, line 83: I think this method can be moved to the parent class with some small modifications. 

For instance, if you create the abstract property `camera`, replace all instances of `self.comcam` by `self.camera`.

Then, you can add additional abstract properties for all special attributes, e.g.;

```
data_query=f"instrument='LSSTComCam' AND"
                        f" detector IN {self.config.detectors} AND exposure IN {exposures}"
```

```
REPO = "/repo/LSSTComCam"
```

you will also need a way to differentiate the archivers:

```
self.comcam.rem.ccarchiver.evt_imageInOODS.aget(timeout=20)
```

could be;

```
self.image_in_oods.aget(timeout=20)
```

where image_in_oods is an abstract property that points to

```
self.comcam.rem.<archiver>.evt_imageInOODS
```

tribeiro (commented):
Here some additional comments. I made mostly comments on the base class and the comcam version of it. I think those will impact all the other classes.

python/lsst/ts/externalscripts/base_make_bias.py
plazas: I agree. There's some code duplication in the `arun` method. For the moment, I would like to make things work properly. I'll see if I can come with a simplification for the next iteration of the review, if not, it will be for future work, if that's OK.

python/lsst/ts/externalscripts/maintel/make_comcam_bias.py
```diff
@@ -0,0 +1,76 @@
...
+        self._instrument_name = 'LSSTComCam'
+
+    @property
+    def camera(self):
+        return self._comcam
+
+    @property
+    def instrument_name(self):
+        """String with instrument name for pipeline task"""
+        return self._instrument_name
```
tribeiro, line 52: consider returning the string `"LSSTComCam"` instead of a private object. This tells readers that this is really a fixed property rather than something that could potentially be updated privately.

python/lsst/ts/externalscripts/auxtel/make_latiss_bias.py
```diff
@@ -0,0 +1,76 @@
...
+        self._instrument_name = 'LATISS'
+
+    @property
+    def camera(self):
+        return self._latiss
+
+    @property
+    def instrument_name(self):
+        """String with instrument name for pipeline task"""
+        return self._instrument_name
```
tribeiro, line 52: Same comment as in the comcam version. Consider returning the string `LATISS` instead of the value of a "private" variable.

python/lsst/ts/externalscripts/base_make_bias.py
```diff
@@ -0,0 +1,227 @@
...
+        for exp in exposures:
+            obs_day = f"{exp}"[:8]
+            temp = int(f"{exp}"[8:])
+            # See example of ack.obsid below
+            seq_num = f"{temp:06}"
+            obs_id = obs_day + "_" + seq_num
+            exposure_set.add(obs_id)
+
+        images_remaining = len(exposure_set)*n_detectors
+        max_counter = 1000
```
tribeiro, line 154: consider making `max_counter` either a class attribute (`self.max_counter=1000` in `__init__`) or making it a configuration parameter....

python/lsst/ts/externalscripts/base_make_bias.py
```diff
@@ -0,0 +1,227 @@
...
+        counter = 0
+        while images_remaining > 0:
+            ack = await self.image_in_oods.next(flush=False, timeout=600)
```
tribeiro, line 157: consider making the timeout(`=600`) a class attribute (`self.oods_timeout=600 ` in `__init__`) or a configuration parameter. 

I think a configuration is best.

python/lsst/ts/externalscripts/base_make_bias.py
```diff
@@ -0,0 +1,227 @@
...
+            # ack.obsid is of the form, e.g., CC_O_20210708_000019
+            if f"{ack.obsid[-15:]}" in exposure_set:
+                images_remaining -= 1
+                counter = 0
+            if counter == max_counter:
+                self.log.info(f"WARN: Maximum number of loops ({max_counter}) reached while waiting "
```
tribeiro, line 163: Should this be `self.log.warning` instead? In any case, you probably want to remove the "WARN" from the message... let the log level speak by itself... :)

python/lsst/ts/externalscripts/base_make_bias.py
```diff
@@ -0,0 +1,227 @@
...
+            # certify the bias
+            REPO = self.config.repo
+            # This is the output collection where the OCPS puts the biases
+            BIAS_DIR = f"u/ocps/{job_id}"
+            CAL_DIR = self.config.calib_dir
+            cmd = (f"butler certify-calibrations {REPO} {BIAS_DIR} {CAL_DIR} " 
+                    "--begin-date 1980-01-01 --end-date 2050-01-01 bias")
+            self.log.info(cmd)
+            try:
+                os.system(cmd)
```
tribeiro, line 220: uhm.. using `os.system` is probably not the best idea here. If this call blocks it will seem like the Script died (it will stop sending heartbeats) and it will also become unresponsive to DDS commands... 

Asyncio has a utility function to run "system calls" without blocking the event loop. 

Use:

```
process = await asyncio.create_subprocess_shell(cmd)

stdout, stderr = await process.communicate()

self.log.debug(f"Process returned: {process.returncode}")
```

you probably want to raise an exception if the return code is different than zero. 

You can also log the output if you want, probably if there's an error:

```
self.log.debug(stdout)
self.log.error(stderr)
```

tribeiro: Another thing to consider here, Is this butler instance accessible at the summit and/or script queue?

plazas: I'm not sure I understand, my apologies. When running this script from a notebook at {{summit-lsp.lsst.codes}} the DM stack needs to be loaded first via something like {{source $LOADSTACK }}, for example, so that {{butler certify}}  is available.

tribeiro (approved):
Looks great! thanks for all the improvements! the use of the base script and the specializations looks really nice!

PR #35, opened 2021-08-23
DM-31306: Create single SAL script that produces biases, darks, and flats 

plazas:


python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,640 @@
...
+
+        return tuple(
+            [
+                (await self.camera.take_imgtype(image_type, exp_time, 1))[0]
+                for exp_time in exp_times
+            ]
+        )
+
+    async def image_in_oods_callback(self, data):
+        """Callback function to check images are in archiver"""
```
czwa, line 311: Docstring for `data`?

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,640 @@
...
+
+        Returns
+        -------
+        response : `dict`
+             Dictionary with the final OCPS status.
+        """
+
+        # Run the pipetask via the OCPS
+        if image_type == "BIAS":
+            pipe_yaml = "cpBias.yaml"
```
czwa, line 373: This is where my comment above about camera specific pipelines would apply.  I think the names can be constructed from the camera's name, but I'm not 100% sure.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,640 @@
...
+        elif image_type == "PTC":
+            pipe_yaml = "measurePhotonTransferCurve.yaml"
+            config_string = (f"-j {self.config.n_processes} -i {self.config.input_collections_ptc} "
+                             f"-i {self.config.calib_collection} "
+                             "--register-dataset-types "
+                             f"{self.config.config_options_ptc}")
+            exposure_ids = exposure_ids_dict["FLAT"]
+        else:
+            raise RuntimeError("Invalid image or calib type {image_type} in 'call_pipetask' function. "
+                               "Valid options: ['BIAS', 'DARK', 'FLAT', 'DEFETCS', 'PTC']")
```
czwa, line 412: Continued typo on DEFECTS.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,640 @@
...
+        exposures : `tuple`
+             Tuple with the IDs of the exposures taken.
+        """
+
+        exp_times = await self.set_exp_times_per_im_type(image_type)
+
+        n_detectors = len(tuple(map(int, self.config.detectors[1:-1].split(','))))
+
+        self.number_of_images_expected = len(exp_times)*n_detectors
+        self.number_of_images_taken = 0
```
czwa, line 336: Is there a check that `number_of_images_taken == number_of_images_expected`?  I don't think I see one here, or below when this is called.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,640 @@
...
+                response_ocps_verify_pipetask = response_ocps_calib_pipetask
+                previous_step = "generation"
+
+            # 4. Certify the calibration
+            if not response_ocps_verify_pipetask['phase'] == 'completed':
+                raise RuntimeError(f"{im_type} {previous_step} not completed successfully: "
+                                   f"Status: {response_ocps_verify_pipetask['phase']}. "
+                                   f"{im_type} certification could not be performed.")
+            else:
+                await self.certify_calib(im_type, job_id_calib)
```
czwa, line 636: This will add a check when we move out of auto-certify mode, correct?  Should that ticket be filed and noted in a comment here so that that doesn't get forgotten?

python/lsst/ts/externalscripts/maintel/make_comcam_calibrations.py
```diff
@@ -0,0 +1,131 @@
...
+        title: MakeComCamCalibrations v1
+        description: Configuration for making a LSSTComCam master calibrations SAL Script.
+        type: object
+        properties:
+            filter:
+                description: Filter name or ID; if omitted the filter is not changed.
+                anyOf:
+                  - type: string
+                  - type: integer
+                    minimun: 1
```
czwa, line 73: Typo.

Can these schemas be derived from a common start (here and Latiss)?  This seems like a lot of code duplication, and if the only difference is the input collections and repo location (which differ only by a fixed string), then there should be a simpler way.

And I think this also answers my question of "where is the filter set", although I don't see where it's sent to the camera (I'm probably just missing something).

python/lsst/ts/externalscripts/auxtel/__init__.py
```diff
@@ -24,3 +24,4 @@
 from .latiss_acquire_and_take_sequence import *
 from .latiss_cwfs_align import *
 from .make_latiss_bias import *
+from .make_latiss_calibrations import *
```
czwa, line 27: This looks like it supersedes the `make_X_bias` code.  Is that correct, and is the plan to have just the main multi-calibration script?

czwa (changes requested):
I'm concerned about amount of code duplication here, and wonder if there's not a simpler way.

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
```diff
@@ -0,0 +1,131 @@
...
+
+class MakeLatissCalibrations(BaseMakeCalibrations):
+    """Class for taking images, constructing, verifying, and
+       certifying master calibrations with LATISS.
+    """
+
+    def __init__(self, index=1):
+        super().__init__(
+            index=index,
+            descr="This class takes biases, darks, and flats  with Auxtel-LATISS, "
```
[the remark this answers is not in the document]
plazas: And I added: "It also optionally produces defects and Photon Transfer Curves."

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,640 @@
...
+                descriptor: Detector IDs.
+            do_verify:
+                type: boolean
+                descriptor: Should the master calibrations be verified? (c.f., cp_verify)
+                default: true
+            config_options_bias:
+                type: string
+                descriptor: Options to be passed to the command-line bias pipetask. They will overwrite \
+                    the values in cpBias.yaml.
+                default: "-c isr:doDefect=False -c isr:doLinearize=False -c isr:doCrosstalk=False \
```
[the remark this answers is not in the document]
plazas: Thanks. I'll use the corresponding instrument pipelines for all the pipetasks.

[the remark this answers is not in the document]
plazas: I don't want to make these parameters mandatory, so I'm going to leave one redundant option as default (unless there's another way).

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,640 @@
...
+            f"n_flat: {config.n_flat}, "
+            f"instrument: {self.instrument_name} "
+        )
+
+        self.config = config
+
+    def set_metadata(self, metadata):
+        """Set estimated duration of the script.
+        """
+        metadata.duration = self.config.n_bias*(self.camera.read_out_time + self.estimated_process_time)
```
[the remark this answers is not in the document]
plazas: I added the number of darks and flats in addition to n_biases as prefactor.

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: I had to do some if/else clauses using the instrument name method, because of different capitalizations in the path names and the instrument names that the pipetask takes (e.g., LATSS vs Latiss and LSSTcomCam vs LsstComCam).

python/lsst/ts/externalscripts/auxtel/__init__.py
plazas: That's correct.

python/lsst/ts/externalscripts/maintel/make_comcam_calibrations.py
plazas: I actually tried to move the redundant lines from the LSSTComCam and LATISS code to the schema in the base class. The only difference would be the instrument name. Unfortunately, there does not seem a way to access the methods in the base class (such as the ``instrument name`` method) from the class method ``get_schema``, so I'm going to have to revert to how things were.

plazas: (to clarify, I tested the changes at the NTS and did not work; and consulted with Tiago to see if there was a way, but it doesn't look like it).

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: Would that be ticket DM-31897?

plazas: I added a reference to that ticket in a comment.

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: I think that's done in the function ``image_in_oods_callback`` define in L313.  it increases a counter every time an image is taken, and then sets ``image_in_oods_received_all_expected`` when the condition is satisfied (i.e., taken == expected).

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
```diff
@@ -0,0 +1,143 @@
...
+        self._latiss = LATISS(domain=self.domain, log=self.log)
+        self._ocps_group = RemoteGroup(domain=self.domain, components=["OCPS:1"], log=self.log)
+
+    @property
+    def camera(self):
+        return self._latiss
+
+    @property
+    def ocps_group(self):
+        # OCPS:1 for LATISS
```
tribeiro, line 53: Consider using a docstring instead of a comment.

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
```diff
@@ -0,0 +1,143 @@
...
+        return self.ocps_group.rem.ocps_1
+
+    @property
+    def instrument_name(self):
+        """String with instrument name for pipeline task"""
+        return "LATISS"
+
+    @property
+    def image_in_oods(self):
+        """Archiver"""
```
tribeiro, line 67: How about, 

```
"""Archiver imageInOODS event."""
```
?

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,659 @@
...
+
+        self.number_of_images_expected = None
+        self.number_of_images_taken = 0
+        self.image_in_oods_samples = dict(BIAS=[], DARK=[], FLAT=[])
+
+        self.number_total_images = None
+
+        self.current_image_type = None
+
+    # Define the OCPS Remote Group (base class) to be able to check
```
tribeiro, line 60: Consider making this a docstring instead of a comment on top of the method.

python/lsst/ts/externalscripts/base_make_calibrations.py
tribeiro: The `image_in_oods_callback` has the logic to handle this condition. Basically if we get less images, the line below:

```
await asyncio.wait_for(self.image_in_oods_received_all_expected.wait(),
                                timeout=self.config.oods_timeout)
```

will timeout. Now, there could be a situation where we get more images than we expect, which probably means someone else commanded the camera to take images. This could be a fail condition which we check afterwards.

python/lsst/ts/externalscripts/maintel/make_comcam_calibrations.py
```diff
@@ -0,0 +1,143 @@
...
+
+class MakeComCamCalibrations(BaseMakeCalibrations):
+    """Class for taking images, constructing, verifying, and
+       certifying master calibrations with LSSTComCam.
+    """
+
+    def __init__(self, index=1):
+        super().__init__(
+            index=index,
+            descr="This class takes bias, darks, and flat exposures with LSSTComCam, "
```
tribeiro, line 38: See my comments about the description in make_latiss_calibrations

python/lsst/ts/externalscripts/maintel/make_comcam_calibrations.py
```diff
@@ -0,0 +1,143 @@
...
+        self._comcam = ComCam(domain=self.domain, log=self.log)
+        self._ocps_group = RemoteGroup(domain=self.domain, components=["OCPS:2"], log=self.log)
+
+    @property
+    def camera(self):
+        return self._comcam
+
+    @property
+    def ocps_group(self):
+        # OCPS:2 for LSSTComCam
```
tribeiro, line 53: See my comments in make_latiss_calibrations.

python/lsst/ts/externalscripts/maintel/make_comcam_calibrations.py
```diff
@@ -0,0 +1,143 @@
...
+        return self.ocps_group.rem.ocps_2
+
+    @property
+    def instrument_name(self):
+        """String with instrument name for pipeline task"""
+        return "LSSTComCam"
+
+    @property
+    def image_in_oods(self):
+        """Archiver"""
```
tribeiro, line 67: See my comment in make_latiss_calibrations.

tribeiro (approved):
Looks good! I left some comments inline for you to consider. From my side it is good to merge once you take a look at those...

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -0,0 +1,665 @@
...
+            f"instrument: {self.instrument_name} "
+        )
+
+        self.config = config
+
+    def set_metadata(self, metadata):
+        """Set estimated duration of the script.
+        """
+        n_images = self.config.n_bias + self.config.n_dark + self.config.n_flat
+        metadata.duration = n_images*(self.camera.read_out_time + self.estimated_process_time)
```
czwa, line 291: Is this `metadata.duration` an estimate that causes a timeout, or is it just an estimate?  My concern is that if `n_dark` is 10, and the commanded exposure time is 100s, this will underestimate the real time.  Similarly the flats have non-zero exposure time.

czwa (approved):
I think I'm happy, just one question on the `metadata.duration`.

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: I don't think it causes a timeout, it's just an estimate. Is that correct, @tribeiro ?

PR #37, opened 2021-09-13
Tickets/dm 31582

patrickingraham:


python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -588,9 +595,9 @@ def get_schema(cls):
                 type: number
                 default: 0.8
               dataPath:
-                description: Path to the butler data repository.
+                description: Path to the butler data repository on the summit
```
tribeiro, line 598: Maybe mention that this is gen3? Also is it "on the summit" or "at the summit"?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -373,10 +373,17 @@ async def run_cwfs(self):
             self.log.info(
                 f"Running cwfs in " f"{self.intra_visit_id}/{self.extra_visit_id}."
             )
+            self.log.info(
```
tribeiro, line 376: Consider using debug level for these messages.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -219,7 +219,7 @@ def get_schema(cls):
               dataPath:
                 description: Path to the butler data repository.
```
tribeiro, line 220: Maybe mention this is gen3?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: I'd say "on" .... although I think either works...

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: I changed a bunch of them throughout and nearby... but actually left this specific instance.

tribeiro (approved):
Looks good!

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
mfisherlevine: In general, we're trying not to do this anywhere - once Gen3 goes (which will be quite soon now, really) then Gen3 will be the only butler in town. I know this has merged, but just a general not to try not to do this if you can help it.

PR #38, opened 2021-09-22
DM-31755: Add slew to a target capabilities in latiss_cwfs_align

tribeiro:


python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -574,6 +573,64 @@ def get_schema(cls):
...
+                  radius:
+                    type: number
+                    description: Radius of the cone search (in degrees).
+              track_target:
+                type: object
+                additionalProperties: false
+                required:
+                  - target_name
+                description: >-
+                    Optional configuration section. Track a specified targer to
```
patrickingraham, line 610: target not targer

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -637,6 +694,30 @@ async def configure(self, config):
...
+            self.cwfs_target_dec = None
+            if "icrs" in config.track_target:
+                self.cwfs_target_ra = config.track_target["icrs"]["ra"]
+                self.cwfs_target_dec = config.track_target["icrs"]["dec"]
+        else:
+            self.log.debug("No target configured.")
+            self.cwfs_target = None
+            self.cwfs_target_ra = None
+            self.cwfs_target_dec = None
+
```
patrickingraham, line 720: Is it worth putting a checkpoint here? 
I admit I'm a little lost on how checkpoints will be used in the future. Maybe after the slew is more appropriate.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -667,8 +748,22 @@ def set_metadata(self, metadata):
         metadata.filter = f"{self.filter},{self.grating}"
 
     async def arun(self, checkpoint=False):
-        """Perform the loop to perform CWFS measurements and hexapod
-        adjustments until the threshold is reached."""
+        """Perform CWFS measurements and hexapod adjustments until the
+        threshold is reached.
```
patrickingraham, line 752: ...until the thresholds are reached.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -667,8 +748,22 @@ def set_metadata(self, metadata):
...
+        """
+
+        if self.cwfs_target is not None and self.cwfs_target_dec is None:
+            await self.slew_object(
+                name=self.cwfs_target, rot=0.0, rot_type=RotType.PhysicalSky
+            )
+        elif self.cwfs_target is not None and self.cwfs_target_dec is not None:
```
patrickingraham, line 759: If there is no ra specified, does slew_icrs give a coherent error message saying so?

patrickingraham (approved):
Some minor comments, please check before merging.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
tribeiro: I think I see your point... Here I am assuming the script was configured using the configure method which will never yield dec without ra. But I imagine one could try to run it without going over configure. I am actually not sure what would happen in it ra is None, so I'll protect against it.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
tribeiro: Can't have checkpoints in `configure`... but I'll add a checkpoint after it slew...

PR #41, opened 2021-10-13
DM-31936: Have an Observing Specialist successfully run OCPS script from DM-31306 

plazas:
Check for camera-specific yaml file; if not found, use the camera-agnostic pipeline file.

### DIFF
--- a/python/lsst/ts/externalscripts/base_make_calibrations.py
+++ b/python/lsst/ts/externalscripts/base_make_calibrations.py
@@ -456,16 +456,21 @@ class BaseMakeCalibrations(salobj.BaseScript, metaclass=abc.ABCMeta):
         # Use the camera-agnostic yaml file if the camera-specific
         # file does not exist.
         cp_pipe_dir = getPackageDir("cp_pipe")
-        pipeline_yaml_file = os.path.join(cp_pipe_dir, "pipelines", pipeline_instrument, pipe_yaml)
+        pipeline_yaml_file = os.path.join(
+            cp_pipe_dir, "pipelines", pipeline_instrument, pipe_yaml
+        )
         file_exists = os.path.exists(pipeline_yaml_file)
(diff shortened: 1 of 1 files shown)


python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -452,10 +453,19 @@ async def call_pipetask(self, image_type, exposure_ids_dict):
         else:
             raise RuntimeError("Nonvalid instrument name: {self.instrument_name")
 
+        # Use the camera-agnostic yaml file if the camera-specific
+        # file does not exist.
+        cp_pipe_dir = os.environ.get("CP_PIPE_DIR")
+        pipeline_yaml_file = f"{cp_pipe_dir}/pipelines/"+f"{pipeline_instrument}/{pipe_yaml}"
```
czwa, line 459: This might be better using `lsst.utils.getPackageDir` and `os.path.join`:
```
from lsst.utils import getPackageDir
cp_pipe_dir = getPackageDir("cp_pipe")
pipeline_yaml_file = os.path.join(cp_pipe_dir, "pipelines", pipeline_instrument, pipe_yaml)
```

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -452,10 +453,19 @@ async def call_pipetask(self, image_type, exposure_ids_dict):
...
+        file_exists = os.path.exists(pipeline_yaml_file)
+        if file_exists:
+            pipeline_yaml_file = "${CP_PIPE_DIR}/pipelines/"+f"{pipeline_instrument}/{pipe_yaml}"
```
czwa, line 462: If the file exists, you don't need to reset it here, do you?  Also, I think this has retained bash style variables that I don't think will work.

PR #42, opened 2021-10-20
DM-31897: Have OCPS scripts use cp_verify output

plazas:


python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+            Dictionary reporting the different thresholds used to decide
+            whether a calibration should be certified or not (see `Notes`
+            below). If there are not any tests that failed,
+            `None` will be returned.
+
+        verify_stats : `dict`
+            Statistics from cp_verify.
+        """
+        # Collection name containing the verification outputs.
+        verifyCollection = f"u/ocps/{job_id_verify}"
```
tribeiro, line 645: Mixing `camelCase` with `snake_case` in the code. I'd suggest `verify_collection`.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+        # Collection that the calibration was constructed in.
+        genCollection = f"u/ocps/{job_id_calib}"
```
tribeiro, line 647: `gen_collection`

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+        numStatErrors : `dict`[`str`][`str`]
+            Dictionary with the total number of tests failed per exposure and
+            per cp_verify test type. If there are not any tests that failed,
+            `None` will be returned.
+
+        thresholds : `dict`[`str`][`int`]
+            Dictionary reporting the different thresholds used to decide
+            whether a calibration should be certified or not (see `Notes`
+            below). If there are not any tests that failed,
+            `None` will be returned.
```
tribeiro, line 639: See my comment above. 

If you are returning `None`, please, update the description to reflect that, 

```
thresholds : `dict`[`str`][`int`] or `None`
```

Although, I would advocate against returning `None` in case of success. I'd rather  return a dictionary indicating that 0 tests failed.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+
+        Returns
+        -------
+        certify_calib : `bool`
+            Booolean indicating if the calibration should be certified or not.
+
+        numStatErrors : `dict`[`str`][`str`]
+            Dictionary with the total number of tests failed per exposure and
+            per cp_verify test type. If there are not any tests that failed,
+            `None` will be returned.
```
tribeiro, line 633: If you are returning None, please, update the description to reflect that,

```
numStatErrors : `dict`[`str`][`int`] or `None`
```

Although, I would strongly advocate against returning `None` in case of success. I'd rather return a dictionary indicating that 0 tests failed.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
 
         return response
 
+    async def check_verification_stats(self, image_type, job_id_calib, job_id_verify):
```
tribeiro, line 608: General comment regarding the return value of this method. Wouldn't it be best to return a single dictionary with all the information stored in it?

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+
+        Returns
+        -------
+        certify_calib : `bool`
+            Boolean assessing whether the calibration should be certified.
+
+        total_counter_failed_tests : `dict`[`str`][`str`]
+            Dictionary with the total number of tests failed per exposure and
+            per cp_verify test type. If there are not any tests that failed,
+            `None` will be returned.
```
tribeiro, line 693: See my previous comments about return `None`.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+                        counter[test] += 1
+                    else:
+                        counter[test] = 1
+                total_counter_failed_tests[exposure] = counter
+            else:
+                continue
+
+        # If there are not exposures with tests that failed.
+        if len(total_counter_failed_tests) == 0:
+            return certify_calib, None
```
tribeiro, line 753: This is returning 2 values, shouldn't it be 3?

```
return certify_calib, None, None
```

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+        -------
+        certify_calib : `bool`
+            Boolean assessing whether the calibration should be certified.
+
+        total_counter_failed_tests : `dict`[`str`][`str`]
+            Dictionary with the total number of tests failed per exposure and
+            per cp_verify test type. If there are not any tests that failed,
+            `None` will be returned.
+
+        thresholds : `dict`[`str`][`int`]
```
tribeiro, line 695: Is this one supposed to be None if it fails? I see there is a return of 2 values if it succeeds and 3 values if it fails. I think that will be problematic to unpack.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
...
+        # Define failure threshold per exposure
+        failure_threshold_exposure = (
+            min_number_failures_per_detector_per_test * min_number_failed_detectors
+        )
+
+        # Count the number of failures per test per exposure.
+        total_counter_failed_tests = {}
+        # Pop this key so that we are left with
+        # exposure ID's as keys only.
+        verify_stats.pop("SUCCESS")
```
tribeiro, line 735: This line is changing the input dictionary, are you sure you want to do that? You may want to create a local copy of it to operate if you intend on doing this. But you could also do something like:

```
for exposure in [key for key in verify_stats if key != "SUCCESS"]:
```

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -722,19 +908,60 @@ async def arun(self, checkpoint=False):
...
                     f"Status: {response_ocps_verify_pipetask['phase']}. "
                     f"{im_type} certification could not be performed."
                 )
             else:
-                await self.certify_calib(im_type, job_id_calib)
+                # Check the verification statistics and decide whether
+                # the master calibration gets certified or not.
+                job_id_verify = response_ocps_verify_pipetask["jobId"]
+
+                # Check verifycation tests only if bias,
```
tribeiro, line 924: I am a bit confused about this statement. Aren't this all the image types this script operates with? Seems a bit redundant to have the if/else statement. But maybe I am missing something.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -722,19 +908,60 @@ async def arun(self, checkpoint=False):
...
+                        )
+                    else:
+                        raise RuntimeError(
+                            f"{im_type} calibration was not certified. \n"
+                            "The number of tests that did not pass per test type per exposure is: "
+                            f"{verify_report} \n"
+                            "Thresholds used to decide whether a calibration should be certified or not: "
+                            f"{thresholds_report} \n"
+                            "MIN_FAILURES_PER_DETECTOR_PER_TEST_TYPE_THRESHOLD is given by the config "
+                            "parameter: 'number_verification_tests_threshold' \n"
```
tribeiro, line 953: Probably a good idea to add the value here as well.

tribeiro (changes requested):
This looks good, I left some general comments in the code. Please take a look..

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -599,6 +605,186 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
 
         return response
 
+    async def check_verification_stats(self, image_type, job_id_calib, job_id_verify):
+        """Check verification statistics.
+
+        Parameters
+        ----------
+        image_type:
+            Image type. Verification currently only implemented for ["BIAS",
```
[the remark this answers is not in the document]
plazas: "Image Type" can actually be `[BIAS, DARK, FLAT, DEFECTS, PTC]`. I can explain that in the documentation, or change the variable name to something like `calibration_name`.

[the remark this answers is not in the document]
plazas: So, in the future, we will have verification for `DEFECTS` and `PTC` too (but we don't have that in this script at this moment).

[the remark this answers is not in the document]
plazas: I clarified the doctring.

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: Yes, sorry. The DM style guide was `camelCase` only until just recently (now there's an option, I think), but sometimes my muscle memory just does it like that and I forget that Telescope and Site is `snake_case` :)

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: `im_type` can be an image or calibration type in `["BIAS", "DARK", "FLAT", "DEFECTS", "PTC"]`, but verification is not yet implemented for `"DEFECTS"` and `"FLAT"`. I added this clarification to the comment.

tribeiro (approved):

PR #43, opened 2021-11-10
Tickets/dm 31995

patrickingraham:


.githooks/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
tribeiro, line 1: I think we stopped using the githook in favor of the pre-commit.. The package should already be configure with it (https://github.com/lsst-ts/ts_externalscripts/blob/develop/.pre-commit-config.yaml) though, I must admit I am not really sure how they work.. maybe @r-owen can wave in here..

tribeiro (approved):
Looks good, but please see my comment about the hook..

.githooks/pre-commit
r-owen: I'm afraid that issue is still present. I wish I'd seen this before I merged my update. I'll make a small branch for this.

r-owen: Good catch. I fixed this (as an extra commit after merging my ticket).

PR #49, opened 2022-03-01
DM-32965: Determine appropriate error message on LOVE when verification fails in OCPS script to make calibrations (fatal errors and warnings)

plazas:


python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -838,21 +838,145 @@ async def count_failed_verification_tests(
...
+            String with full report.
+        """
+
+        verify_report = report_check_verify_stats["NUM_STAT_ERRORS"]
+        thresholds_report = report_check_verify_stats["FAILURE_THRESHOLDS"]
+        verify_stats = report_check_verify_stats["VERIFY_STATS"]
+
+        final_report_string = ""
+        # List exposure IDs that have tests that failed
+        final_report_string += "|Exposures with verification tests that failed:\n"
```
tribeiro, line 894: Oh, I see here and elsewhere you did `"| <text> \n"`, I think in order for this to work you need to do `"<text> \n|"`.

tribeiro (commented):
So, my only question is about how the wall break workaround got implemented and if it actually worked. Everything else looks good. Let's figure this out first before we merge. Because, if neither workaround works, we should probably remove the "|" altogether

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -838,21 +838,145 @@ async def count_failed_verification_tests(
...
+        final_report_string += "will not be certified. "
+
+        final_report_string += (
+            "In terms of the threshold values, this amounts for the condition that\n|"
+        )
+        final_report_string += (
+            "the final number of exposures that failed verification is greater than\n|"
+        )
+        final_report_string += (
+            "or equal to the acceptable maximum number of failed exposures"
```
r-owen, line 975: Creating a string by repeatedly appending to it is very inefficient.

Please consider accumulating a list of entries and useing join, instead. If `"\n|"` appears before every entry then you can omit it and use it in the join, instead. Here is an example:
```
final_report_strings = [
    "Exposures with verification tests that failed:",
    " ".join(str(exposure) for exposure in verify_report),
    "Number of tests that failed per test type:"
    f"\t| Exposure ID: {exposure}",
]
for test_type in verify_report[exposure]:
    final_report_strings.append(
        f"\t|{test_type}: {verify_report[exposure][test_type]}"
    )
final_report_strings += [
    "Test types that failed verification per exposure,",
    "detector, and amplifier:",
    ...
]
...
final_report_string = "\n|".join(final_report_strings)
```

And any "line" can be as many lines of code as you want, e.g. if the list of exposures should appear right after "Exposures with..." then just delete the comma separating those two entries to concatenate those two lines, or add a `+` to make it a bit more obvious that the concatenation is intentional:
```
final_report_strings = [
    "Exposures with verification tests that failed: "
    + " ".join(str(exposure) for exposure in verify_report),
```

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1036,45 +1159,59 @@ async def arun(self, checkpoint=False):
...
+                                    await self.build_verification_report_summary(
+                                        report_check_verify_stats
+                                    )
+                                )
+                                self.log.error(final_report_string)
+                                self.log.error(
+                                    f"Generation collection: {gen_collection}"
+                                )
+                                self.log.error(
+                                    f"Verification collection: {verify_collection}"
```
r-owen, line 1197: Are you sure you want these to be 3 different log messages? I'm a bit uneasy about that because they apparently relate to each other, but once they are published as 3 different messages that becomes harder to reconstruct later.

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: I'll be removing the "|"; we determined with Tiago that they don't work.

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: I'll unify the messages

PR #51, opened 2022-03-17
Force extrafocal image to use intrafocal location on collision

mfisherlevine:


tribeiro:
I think we can close this PR right?

mfisherlevine:
Yup, looks like this fix got translated and merged 👍

PR #54, opened 2022-04-01
DM-34181: Running OCPS ComCam Calibrations (3/24/2022)

plazas:


tribeiro (changes requested):
Please, update your development environment black to 22 and reformat the code. You should probably patch up the git history as well, because there is a commit with a black update that is affecting other files.

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
```diff
@@ -86,6 +86,17 @@ def get_schema(cls):
         description: Configuration for making a LATISS calibrations SAL Script.
         type: object
         properties:
+            detectors:
+                description: Detector IDs.
+                type: array
+                items:
+                  - type: number
```
tribeiro, line 93: number is either int or float... I think detector numbers are only integer no?

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
```diff
@@ -86,6 +86,17 @@ def get_schema(cls):
...
+                minContains: 0
+                maxContains: 0
+                minItems: 0
+                maxItems: 1
```
tribeiro, line 97: uhm.. do you really need this to be an array if it is only one item? Do you really need this at all for LATISS?

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -26,6 +26,8 @@
 import asyncio
 import collections
 import os
+import time
+import numpy as np
```
tribeiro, line 30: Please, add an empty line between these import statements

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -296,17 +311,47 @@ async def configure(self, config):
...
+            f"instrument: {self.instrument_name}, "
+            f"script_mode: {config.script_mode}, "
+            f"generate_calibrations: {config.generate_calibrations}, "
+            f"do_verify: {config.do_verify}, "
+            f"do_defects: {config.do_defects}, "
+            f"do_ptc: {config.do_ptc}, "
+            f"do_gain_from_flat_pairs: {config.do_gain_from_flat_pairs} "
         )
 
+        if self.instrument_name == "LATISS":
```
tribeiro, line 334: I don't think this if/else statement should be here. It is kind of doing an inline specialization in the base class.

I don't think LATISS need to deal with detectors at all.

If this is something you need to handle on the parent class, then make an abstract method that is called to handle it and implement if on the child classes.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -296,17 +311,47 @@ async def configure(self, config):
...
+                for x in config.detectors:
+                    d += f"{x},"
+                det_tuple_string = d[:-1] + ")"
+                config.detectors = det_tuple_string
+            else:
+                # Default value is an empty array: use all detectors.
+                config.detectors = "(0..9)"
+                n_detectors = 9
+        else:
+            raise RuntimeError("Incorrect instrument name: {self.instrument_name}.")
```
tribeiro, line 351: yeah, so this becomes immediately obsolete once we have LSSTCam..

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -380,10 +423,31 @@ async def take_images(self, image_type):
...
-            self.image_in_oods_received_all_expected.wait(),
-            timeout=self.config.oods_timeout,
-        )
+        # Wait for image in OODS. Try several times before
+        # giving up, if there's a timeout.
+        do_wait_for_image_in_oods = True
+        retry = 0
+        retry_rate = self.config.oods_timeout_retry_rate
+        max_retry = self.config.oods_timeout_max_retry
+        while do_wait_for_image_in_oods:
```
tribeiro, line 432: Uhm.. so you are doing a "timeout loop" to "timeout" waiting for the events to arrive. I don't think this is a good idea. 

Instead, consider improving the error message such that it tells us what we are missing. 

For example, see what I did in this ticket branch that helped me figure out the problem:

https://github.com/lsst-ts/ts_externalscripts/blob/aabbb62e04a254643e527ca0c34425372a3cadca/python/lsst/ts/externalscripts/base_make_calibrations.py#L404-L423

tests/auxtel/test_make_latiss_calibrations.py
```diff
@@ -51,7 +51,7 @@ async def test_configure(self):
             exp_times_dark = 10
             n_flat = 4
             exp_times_flat = [10, 10, 50, 50]
-            detectors = "(0, 1, 2, 3, 4, 5, 6, 7, 8)"
+            detectors = "[0, 1, 2, 3, 4, 5, 6, 7, 8]"
```
tribeiro, line 54: Is this supposed to be an array or a string? Also, gain, doesn't make sense to have this in LATISS

tribeiro (changes requested):
Made inline comments, please take a look

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
plazas: Yes. I had ``integer`` before in there because of the reason you cite, but that was breaking the schema. Should it be ``int`` instead?

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: Is there a guideline somewhere as to why it should be like this? (I added the line between "time" and "numpy" imports, I'm just curious).

python/lsst/ts/externalscripts/base_make_calibrations.py
plazas: I do need the detectors for the pipetasks. I'll add an abstract class as you suggest.

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
tribeiro: should be `integer`: https://json-schema.org/understanding-json-schema/reference/numeric.html#id4

python/lsst/ts/externalscripts/base_make_calibrations.py
tribeiro: that's a good question.. I'm not sure..

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -444,32 +494,46 @@ async def call_pipetask(self, image_type, exposure_ids_dict):
...
-                f"-i {self.config.calib_collection} "
                 "--register-dataset-types "
                 f"{self.config.config_options_ptc}"
             )
             exposure_ids = exposure_ids_dict["FLAT"]
         else:
             raise RuntimeError(
                 "Invalid image or calib type {image_type} in 'call_pipetask' function. "
-                "Valid options: ['BIAS', 'DARK', 'FLAT', 'DEFECTS', 'PTC']"
+                "Valid options: ['BIAS', 'DARK', 'FLAT', 'DEFECTS', 'PTC', 'GAIN']"
```
tribeiro, line 515: Se my comment above about having a class attribute with this information in it. You should not have to be updating text etc every time you add a new image type.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -444,32 +494,46 @@ async def call_pipetask(self, image_type, exposure_ids_dict):
...
+        # If we are using internal calibrations, place the
+        # calibration collection where they were certified
+        # before the input collections of each task
+        # (except for BIAS, which is the first step).
+        if self.config.generate_calibrations and image_type != "BIAS":
+            split_string = config_string.split()
+            # The third entry is where the input collections are.
+            split_string[3] = f"{self.config.calib_collection}," + split_string[3]
+            config_string = " ".join(split_string)
+
```
tribeiro, line 527: I think I mentioned this before but I don't thing this if/else should be here. Again, if you add support to another camera, you would have to change this code here, which defeats the purpose of having a base class etc.. 

Instead make `pipeline_instrument` an abstract class property.

tribeiro (commented):
General comment about the default configuration on `BaseMakeCalibrations`, I see several parameters have default that don't make a lot of sense to me. 

So, I suggest either making them appropriate defaults (such that we would seldom have to set them) or remove the default and make them "required" configuration. The idea of providing default is such that users won't have to set them most of the time. 

For example `n_bias`, `n_dark`, `n_flat` all have default of 1, which are probably never appropriate values. But I see other places with the same problem.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -296,17 +313,36 @@ async def configure(self, config):
         config: `types.SimpleNamespace`
             Configuration data. See `get_schema` for information about data
             structure.
+
+        Raises
+        ------
+            RuntimeError : Incorrect instrument name.
```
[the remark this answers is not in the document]
plazas: yeah, this was from the previous version, and I forgot to update the documentation.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -408,7 +457,10 @@ async def call_pipetask(self, image_type, exposure_ids_dict):
              Dictionary with the final OCPS status.
         """
 
-        # Run the pipetask via the OCPS
+        # Run the pipetasks via the OCPS.
+        # By default, config.generate_calibrations is 'false'
+        # and the necessary calibrations are assumed to be
+        # in the input collections.
 
         if image_type == "BIAS":
```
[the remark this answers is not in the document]
plazas: Thanks for the suggestion.  I'll have to include the input collections and options for each type as parameters of the function since they are different in each {{config_string}}, but that still should be better than how it is right now.  I assume that you would suggest doing something similar in the function {{verify_calib}}, correct?

python/lsst/ts/externalscripts/auxtel/make_latiss_calibrations.py
```diff
@@ -71,6 +73,21 @@ def instrument_name(self):
...
+
+    @property
+    def detectors(self):
+        """Array with detector IDs"""
+        return self._detectors
+
+    @property
+    def n_detectors(self):
+        """Number of detectors"""
+        return self._n_detectors
```
tribeiro, line 89: Consider removing both `self._detectors` and `self._n_detectors`, and just returning the constant values. e.g.;

```
@property
def n_detectors(self):
    return 1

@property
def detectors(self):
    return [0]
```

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -60,6 +62,46 @@ def __init__(self, index, descr):
 
         self.current_image_type = None
 
+        # Supported calibrations types
+        self.supported_calibrations_generation = [
```
tribeiro, line 66: Instead of having both `self.supported_calibrations_generation` and `self.pipetask_parameters`, consider having only `self.pipetask_parameters`. Same for `self.supported_calibrations_verification` and `self.pipetask_parameters_verification`.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1036,169 +1446,241 @@ async def arun(self, checkpoint=False):
...
+        -------
+        int
+            Exposure id in the format YYYYMMDD12345, e.g., 2022040600007.
+        """
+        _, _, i_prefix, i_suffix = obsid.split("_")
+
+        return int((i_prefix + i_suffix[1:]))
+
+    def get_detectors_string(self, detector_array):
+        """Get a detetcor string from a detctor array.
```
tribeiro, line 1659: typos:
  detetcor -> detector
  detctor -> detector

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1036,169 +1446,241 @@ async def arun(self, checkpoint=False):
...
+        Convert a detetcor array of the form [0, 1, 2, ...] into a
```
tribeiro, line 1660: typo:

  detetcor -> detector

tribeiro (approved):
Alright! Thanks for implementing all my suggestions. I have a few more comments inline but I think it is good to go. Just make sure you take a look at the comments before merging!

PR #55, opened 2022-04-18
tickets/SITCOM-275

mareuter:
Remove extra ack in base_make_calibrations.py.

### DIFF
--- a/python/lsst/ts/externalscripts/base_make_calibrations.py
+++ b/python/lsst/ts/externalscripts/base_make_calibrations.py
@@ -485,6 +485,7 @@ class BaseMakeCalibrations(salobj.BaseScript, metaclass=abc.ABCMeta):
         else:
             pipeline_yaml_file = f"${{CP_PIPE_DIR}}/pipelines/{pipe_yaml}"
 
+        # This returns the in-progress acknowledgement with the job identifier
         ack = await self.ocps.cmd_execute.set_start(
             wait_done=False,
             pipeline=f"{pipeline_yaml_file}",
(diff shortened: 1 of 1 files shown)


python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -493,11 +494,6 @@ async def call_pipetask(self, image_type, exposure_ids_dict):
             data_query=f"instrument='{self.instrument_name}' AND"
             f" detector IN {self.config.detectors} AND exposure IN {exposure_ids}",
         )
-        if ack.ack != salobj.SalRetCode.CMD_ACK:
-            ack.print_vars()
-
-        # Wait for the in-progress acknowledgement with the job identifier.
-        ack = await self.ocps.cmd_execute.next_ackcmd(ack, wait_done=False)
         self.log.debug(
```
tribeiro, line 497: I think we should probably remove the following debug statement...

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -619,11 +615,6 @@ async def verify_calib(self, image_type, job_id_calib, exposure_ids_dict):
             data_query=f"instrument='{self.instrument_name}' AND"
             f" detector IN {self.config.detectors} AND exposure IN {exposure_ids}",
         )
-
-        if ack.ack != salobj.SalRetCode.CMD_ACK:
-            ack.print_vars()
-
-        ack = await self.ocps.cmd_execute.next_ackcmd(ack, wait_done=False)
         self.log.debug(
```
tribeiro, line 618: Another debug to be removed?

tribeiro (approved):
Looks good, I made a few comments about some debug messages left over. I think they should be removed.

python/lsst/ts/externalscripts/base_make_calibrations.py
mareuter: I'm not sure about removing them. They were after the original in-progress ack retrieval so I think leaving them there in the new context still allows one to diagnose a problem if something goes wrong with the command exec. That's how I was able to determine what broke when I went to fix this.

PR #56, opened 2022-04-22
replaced rapid_analysis in latiss scripts and tests.

patrickingraham:
Improved mock of ready_to_take_data
All tests passing on summit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ pytest_session.txt
 .cache/
 .pytest_cache
 .coverage
+*__pycache__*
 
 # vim
 *.swp
--- a/python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
+++ b/python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
@@ -21,7 +21,7 @@
 __all__ = ["LatissAcquireAndTakeSequence"]
 
 import asyncio
-import collections.abc
+import collections
 import warnings
 
 import numpy as np
@@ -34,6 +34,7 @@ from lsst.ts.observatory.control.auxtel import ATCS, LATISS
 from lsst.ts.observatory.control.constants import latiss_constants
 from lsst.ts.observatory.control.utils import RotType
 from lsst.ts.standardscripts.utils import format_as_list
+from lsst.ts.idl.enums.ATPtg import WrapStrategy
 
 try:
     from lsst.ts.observing.utilities.auxtel.latiss.utils import (
@@ -41,7 +42,7 @@ try:
         parse_obs_id,
     )
     from lsst.pipe.tasks.quickFrameMeasurement import QuickFrameMeasurementTask
-    from lsst.rapid.analysis import BestEffortIsr
+    from lsst.summit.utils import BestEffortIsr
 
     from lsst.ts.observing.utilities.auxtel.latiss.getters import get_image
 except ImportError:
@@ -214,6 +215,17 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                     minimum: 0
                 default: 2.
 
+              time_on_target:
+                description: Override the automatically calculated time_on_target value used
+                    to optimize the telescope cable wrap positions.
+                    Set to null (default) to have it calculated automatically.
+
+                anyOf:
+                  - type: "null"
+                  - type: number
+                    minimum: 0.0
+                default: null
+
               datapath:
                 description: Path to the gen3 butler data repository. The default is for the summit.
                 type: string
@@ -313,7 +325,7 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
         # make a list of tuples from the filter, exptime and grating lists
         _recurrences = (
             len(config.exposure_time_sequence)
-            if isinstance(config.exposure_time_sequence, collections.abc.Iterable)
+            if isinstance(config.exposure_time_sequence, collections.Iterable)
             else 1
         )
 
@@ -346,6 +358,41 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                 DeprecationWarning,
             )
 
+        self.time_on_target = config.time_on_target
+        self.time_on_target = (
+            self.compute_time_on_target()
+            if config.time_on_target is None
+            else config.time_on_target
+        )
+
+    def compute_time_on_target(self):
+        """Determine the amount of time spent on target so the rotator position
+        can be optimized. Assumed slew time is 3 minutes, and that the read,
+        calculation, and offset totals 20s.
+
+        Returns
+        -------
+        time_on_target: float
+            Estimate of time on target in seconds.
+        """
+
+        time_on_target = 180 + self.max_acq_iter * (self.acq_exposure_time + 20)
+        if self.do_take_sequence:
+            nexp = len(self.visit_configs)
+            visit_exptime_total = 0.0
+            for (filt, exptime, grating) in self.visit_configs:
+                visit_exptime_total += exptime
+
+            time_on_target += visit_exptime_total
+            # assume 5s for readout and instrument reconfiguration
+            # for each exposure
+            time_on_target += nexp * 5
+        self.log.debug(
+            f"Assuming a time_on_target of {time_on_target/60:0.2f} minutes."
+        )
+
+        return time_on_target
+
     def get_best_effort_isr(self):
         # Isolate the BestEffortIsr class so it can be mocked
         # in unit tests
@@ -356,9 +403,9 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
     # I'm not quite sure what the metadata.filter bit is used for...
     def set_metadata(self, metadata):
         metadata.duration = 300
-        filters, gratings, expTimeTotal = set(), set(), 0
-        for (filt, expTime, grating) in self.visit_configs:
-            expTimeTotal += expTime
+        filters, gratings, exptime_total = set(), set(), 0
+        for (filt, exptime, grating) in self.visit_configs:
+            exptime_total += exptime
             filters.add(filt)
             gratings.add(grating)
         metadata.filter = f"{filters},{gratings}"
@@ -423,6 +470,8 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                 target_name=self.object_name,
                 rot_type=RotType.Parallactic,
                 slew_timeout=240,
+                az_wrap_strategy=WrapStrategy.OPTIMIZE,
+                time_on_target=self.time_on_target,
                 offset_x=dx_arcsec,
                 offset_y=dy_arcsec,
             )
@@ -432,6 +481,8 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
                 name=self.object_name,
                 rot_type=RotType.Parallactic,
                 slew_timeout=240,
+                az_wrap_strategy=WrapStrategy.OPTIMIZE,
+                time_on_target=self.time_on_target,
                 offset_x=dx_arcsec,
                 offset_y=dy_arcsec,
             )
@@ -608,7 +659,7 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
 
         nexp = len(self.visit_configs)
 
-        for i, (filt, expTime, grating) in enumerate(self.visit_configs):
+        for i, (filt, exptime, grating) in enumerate(self.visit_configs):
 
             # Check if a manual focus offset is required
             if self.manual_focus_offset != 0.0 and not self.manual_focus_offset_applied:
@@ -626,7 +677,7 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
 
             # Take an image
             await self.latiss.take_object(
-                exptime=expTime,
+                exptime=exptime,
                 n=1,
                 filter=filt,
                 grating=grating,
@@ -636,7 +687,7 @@ class LatissAcquireAndTakeSequence(salobj.BaseScript):
             )
 
             self.log.info(
-                f"Completed exposure {i + 1} of {nexp}. Exptime = {expTime:6.1f}s,"
+                f"Completed exposure {i + 1} of {nexp}. Exptime = {exptime:6.1f}s,"
                 f" filter={filt}, grating={grating})"
             )
 
--- a/python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
+++ b/python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
@@ -34,13 +34,14 @@ from lsst.ts import salobj
 from lsst.ts.observatory.control.utils import RotType
 from lsst.ts.observatory.control.auxtel.atcs import ATCS
 from lsst.ts.observatory.control.auxtel.latiss import LATISS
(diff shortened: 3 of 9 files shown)


python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -791,6 +806,9 @@ async def configure(self, config):
 
         self.take_detection_image = config.take_detection_image
 
+        # Assume the time-on-target is 10 minutes
+        self.time_on_target = 10 * 60
```
tribeiro, line 810: can you do `self.time_on_target = 600.0` instead?  

Also, doesn't seem like this Is a configuration parameter. Maybe move the definition to the class `__init__` method?

tribeiro (commented):
I made a few comments inline I hope we you can take a look at before merging. The change in the gitignore file is really odd. It is marking the entire file as changed but there's only a single line that got updated. Can you check what's going on?

.gitignore
```diff
@@ -1,21 +1,22 @@
...
-# Pytest
-tests/.tests
-pytest_session.txt
-.cache/
-.pytest_cache
-.coverage
-
-# vim
-*.swp
+.sconsign.dblite
```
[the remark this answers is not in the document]
patrickingraham: ... I don't... but I updated this with sourcetree rather than manually... i wonder if it changed the permissions, or line endings... I can try to change it back.

tests/auxtel/test_latiss_acquire_and_take_sequence.py
```diff
@@ -295,12 +302,12 @@ async def test_configure(self):
 
             acq_filter = "acqfilter"
             acq_grating = "acqgrating"
-            acq_exposure_time = 1
+            acq_exposure_time = 10
```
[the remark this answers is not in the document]
patrickingraham: so this is just a test to see if it configures. The reason for increasing the time was to check that the time_on_target calculation included these values. When they were really small the values were dominated by overhead.
It never actually does the acquisition, therefore it doesn't increase the time.

tests/auxtel/test_latiss_acquire_and_take_sequence.py
```diff
@@ -295,12 +302,12 @@ async def test_configure(self):
...
             acq_filter = "acqfilter"
             acq_grating = "acqgrating"
-            acq_exposure_time = 1
+            acq_exposure_time = 10
             max_acq_iter = 3
             target_pointing_tolerance = 3
             filter_sequence = ["test_filt1", "test_filt2"]
             grating_sequence = "test_disp1"
-            exposure_time_sequence = [1.0, 2.0]
+            exposure_time_sequence = [20.0, 60.0]
```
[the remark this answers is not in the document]
patrickingraham: see above

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +347,26 @@ async def configure(self, config):
                 DeprecationWarning,
             )
 
+        # Determine the amount of time spent on target so the rotator position
+        # can be optimized. Assumed slew time is 3 minutes,
+        # read+calculation+offset is 20s
+        if self.do_acquire:
+
```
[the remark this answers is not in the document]
patrickingraham: I did a few things:
1. Added it as a config parameter... but check the default value handling... i don't like it but don't know how to use "None" as a default value since it isn't a number.
2. made the calculation a method... i suppose it's cleaner but it's not really decreasing the amount of code etc.

tests/auxtel/test_latiss_acquire_and_take_sequence.py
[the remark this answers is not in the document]
tribeiro: makes sense

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -436,9 +437,17 @@ async def run_cwfs(self):
...
-                "Intra and Extra source finding algorithm " "found different sources."
+            self.log.warning(
+                "Source finding algorithm found different sources for intra/extra. \n"
+                f"intra found [y,x] = [{self.intra_result.brightestObjCentroidCofM[1]},"
+                "{self.intra_result.brightestObjCentroidCofM[0]}]\n"
+                f"extra found [y,x] = [{self.extra_result.brightestObjCentroidCofM[1]},"
+                "{self.extra_result.brightestObjCentroidCofM[0]}]\n"
+                "Forcing them to use the intra-location."
             )
+            self.force_extra_focal_box_location = True
```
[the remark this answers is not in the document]
patrickingraham: ok, I like `extra_focal_position_out_of_range`

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -479,18 +488,24 @@ def create_donut_stamps_for_cwfs(self):
             ceny - self.side : ceny + self.side, cenx - self.side : cenx + self.side
         ]
 
-        ceny, cenx = int(self.extra_result.brightestObjCentroidCofM[1]), int(
-            self.extra_result.brightestObjCentroidCofM[0]
-        )
+        if self.force_extra_focal_box_location:
```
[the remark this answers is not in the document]
patrickingraham: I guess I don't see how this really cleans it up... but I implemented it anyways...

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +354,30 @@ async def configure(self, config):
                 DeprecationWarning,
             )
 
+        self.time_on_target = config.time_on_target
+        if config.time_on_target == 0:
```
tribeiro, line 358: if you agree to make `config.time_on_target = None` be the guard value, replace this by:

```
if config.time_on_target is None:
```

you can also do this in 1 line with:

```
self.time_on_target = self.compute_time_on_target() if config.time_on_target is None else config.time_on_target
```

Note it's dealing with `config.time_on_target = None` first, which I guess is the most common occurrence and hence, would make this more efficient.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +354,30 @@ async def configure(self, config):
...
+            self.time_on_target = self.compute_time_on_target()
+
+    def compute_time_on_target(self):
+        """Determine the amount of time spent on target so the rotator position
+        can be optimized. Assumed slew time is 3 minutes, and that the read,
+        calculation, and offset totals 20s."""
```
tribeiro, line 364: Do you want to document that it returns a float?

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +354,30 @@ async def configure(self, config):
...
+
+        self.time_on_target = 180 + self.max_acq_iter * (self.acq_exposure_time + 20)
```
tribeiro, line 366: Consider replacing this variable here with a local variable instead of the class attribute one. Basically, inside this method replace `self.time_on_target` with just `time_on_target`

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -462,35 +477,67 @@ async def run_cwfs(self):
...
+        print(intra_box[0])
         intra_square = self.intra_exposure.image.array[
-            ceny - self.side : ceny + self.side, cenx - self.side : cenx + self.side
+            intra_box[0] : intra_box[1], intra_box[2] : intra_box[3]
         ]
 
-        ceny, cenx = int(self.extra_result.brightestObjCentroidCofM[1]), int(
-            self.extra_result.brightestObjCentroidCofM[0]
-        )
+        if self.extra_focal_position_out_of_range:
```
tribeiro, line 522: This if statement here needs to go to `get_extra_donut_center`. Here I'd expect to see only

```
extra_box = self.get_donut_region(self.get_extra_donut_center())
extra_square = self.extra_exposure.image.array[
extra_box[0] : extra_box[1], extra_box[2] : extra_box[3]
             ]
```

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -462,35 +477,67 @@ async def run_cwfs(self):
         results_dict = self.calculate_results()
         return results_dict
 
+    def get_donut_region(self, center):
```
tribeiro, line 480: Consider making this receive `center_x` and `center_y` so you don't have to create the local variables... 

When calling with `get_intra_donut_center` you can do

```
self.get_donut_region(*self.get_intra_donut_center())
```

same for extra

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -462,35 +477,67 @@ async def run_cwfs(self):
...
+            center_x + self.side,
+        )
+
+    def get_intra_donut_center(self):
+        return [
+            int(self.intra_result.brightestObjCentroidCofM[1]),
+            int(self.intra_result.brightestObjCentroidCofM[0]),
+        ]
+
+    def get_extra_donut_center(self):
```
tribeiro, line 497: Just to clarify, you can make this method something like this:

```
    def get_extra_donut_center(self):
        return (
            int(self.extra_result.brightestObjCentroidCofM[1]),
            int(self.extra_result.brightestObjCentroidCofM[0])
            if not self.extra_focal_position_out_of_range
            else self.get_intra_donut_center(),
        )
```

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +357,40 @@ async def configure(self, config):
...
+        )
+
+    def compute_time_on_target(self):
+        """Determine the amount of time spent on target so the rotator position
+        can be optimized. Assumed slew time is 3 minutes, and that the read,
+        calculation, and offset totals 20s.
+
+        Returns
+        -------
+        self.time_on_target: float
```
tribeiro, line 374: remove the `self` here.. :)

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +357,40 @@ async def configure(self, config):
...
+            Estimate of time on target in seconds.
+        """
+
+        time_on_target = 180 + self.max_acq_iter * (self.acq_exposure_time + 20)
+        if self.do_take_sequence:
+            nexp = len(self.visit_configs)
+            sci_expTimeTotal = 0.0
```
tribeiro, line 381: mix of `snake_case` and `camelCase`.. replace with `sci_exp_time_total` or consider renaming to something like `exp_time_total` or `exp_time_sci_total`...

is "sci" short of science? I think you can remove it...

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +357,40 @@ async def configure(self, config):
...
+            for (filt, expTime, grating) in self.visit_configs:
+                sci_expTimeTotal += expTime
+                time_on_target += sci_expTimeTotal
+
+            # assume 5s for readout and instrument reconfiguration
+            time_on_target += nexp * 5
```
tribeiro, line 387: The comment says, "assume 5s for readout and instrument reconfiguration", but this is multiplying the time by 5.. do you want to add instead? 

Also, maybe make is a class attribute? `instrument_setup_overhead`? you can also use `self.latiss.read_out_time`, but since you are trying to account for more than just the readout time, maybe adding the new "overhead" attribute would be better.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -346,6 +357,40 @@ async def configure(self, config):
...
+        -------
+        self.time_on_target: float
+            Estimate of time on target in seconds.
+        """
+
+        time_on_target = 180 + self.max_acq_iter * (self.acq_exposure_time + 20)
+        if self.do_take_sequence:
+            nexp = len(self.visit_configs)
+            sci_expTimeTotal = 0.0
+            for (filt, expTime, grating) in self.visit_configs:
```
tribeiro, line 382: `camelCase` in `expTime`.. use `exp_time`

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -214,6 +215,16 @@ def get_schema(cls):
                     minimum: 0
                 default: 2.
 
+              time_on_target:
+                description: Override the automatically calculated time_on_target value used
+                    to optimize the telescope cable wrap positions.
+                    Set to zero (default) to calculate automatically.
```
tribeiro, line 221: Update the description.

tribeiro (commented):
just a few more comments.. I think we'll be good after these

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: No, this is adding 5s per exposure, it's just wag but should be small in the grand scheme of things. I added brackets to make it more clear.
I don't think we're doing ourselves any favours by adding class attributes here. This is only a rough estimate.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: used `exptime` as per the rest of the code.
Again, as noted below, this was changed throughout so you'll see many more changes.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: hmm, i commented on this but must have not hit "Comment"
Implemented, but not as a single line... it was too hard to read.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
patrickingraham: it's for the total exposure time for a AuxTel "visit" 

Fixed the camelCase, but it resulted in multiple changes throughout.

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -214,6 +215,16 @@ def get_schema(cls):
...
+                anyOf:
+                  - type: "null"
+                  - type: number
+                    minimum: 0.0
```
tribeiro, line 225: wait, you want the minimum to be null no, so it will compute the value instead of using 0?

PR #57, opened 2022-05-06
DM-34130: Add support for additional image type in observatory control package and document them in a more visible place

tribeiro:
Note that this PR has some updates coming from the AuxTel run branch (DM-34581). I am opening the PR now but I will wait until we get DM-34581 merged than rebase. I just want to get the PR out.


mareuter (approved):
Looks good to me.

PR #61, opened 2022-06-08
DM-35101: OCPS calibration script: failure when "detectors" is omitted from the configuration file

plazas:


tribeiro (approved):
Looks good

PR #65, opened 2022-07-15
Tickets/dm 34991

tribeiro:
The idea of this PR is to add a new script to execute the optical alignment of AuxTel using the wep code we will also use for the main telescope. Because the script is doing the exact same thing as the currently existing `LatissCWFSAlign`, and in order to avoid copy/pasting a bunch of code I decided to create a base class `LatissBaseAlign` that contains all (or most of) the common code. This includes the script configuration, slewing/offsetting/taking data etc. Then there is an abstract method called `run_align` which is in charge of running the wavefront estimation code and outputting a result.

The original class was updated to use the new base class.

A new class `LatissWEPAlign` was added (alongside a script to run it), which implements the wep part.

Some things to note:

1 - This new class runs the code in a separate process, rather than a separate thread. This keeps he main thread free to manage the script. The caveat to that is that we need to isolate all the data going to and coming back from the process. To that, I am not pulling all the data generated on the process back, only what is really needed, otherwise we would loose a lot of time with Process IO.

2 - It is using BestEffortISR to process the data and `QuickFrame` to find the sources, just like the original script. I decided for this venue to reduce the adoption friction, since we have been testing this for quite a while successfully. 

3 - I could not find a good way to reuse the unit tests. So here I had to resort to copy/paste. :(

4 - Speaking of unit tests, they are not passing. The reason, as you saw the discussion on slack, is that the results we get from cwfs and wep are not compatible within the specified tolerances. Then we found the issue with the obscuration in the original version. I guess we can think about what we will do in the future. Note that these tests only run on the test stand and you need to have ts_wep for the wep tests to run. So, the CI tests passes, but only because it is skipping the tests that need data.

### DIFF
--- a/python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
+++ b/python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
@@ -1,0 +1,850 @@
+# This file is part of ts_externalcripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["LatissBaseAlign"]
+
+import abc
+import time
+import types
+import yaml
+import typing
+import dataclasses
+
+import numpy as np
+
+from lsst.ts import salobj
+from lsst.ts.observatory.control.utils import RotType
+from lsst.ts.observatory.control.auxtel.atcs import ATCS
+from lsst.ts.observatory.control.auxtel.latiss import LATISS
+from lsst.ts.idl.enums.ATPtg import WrapStrategy
+
+STD_TIMEOUT = 10
+
+
+@dataclasses.dataclass
+class LatissAlignResults:
+    zernikes: typing.Tuple[float, float, float]
+    zernikes_rot: typing.Tuple[float, float, float]
+    offset_hex: typing.Tuple[float, float, float]
+    offset_tel: typing.Tuple[float, float, float]
+
+
+class LatissBaseAlign(salobj.BaseScript, metaclass=abc.ABCMeta):
+    """Implements generic behavior for scripts that execute curvature wavefront
+    sensing, abstracting the part that performs the measurements.
+
+    This metaclass implements the basic functionality of finding targets and
+    taking the (intra/extra focal) data. Child classes are left to implement
+    the processing of the data and returning the results.
+
+    Parameters
+    ----------
+    index : `int`, optional
+        Index of Script SAL component (default=1).
+    remotes : `bool`, optional
+        Should the remotes be created (default=True)? For unit testing this
+        can be set to False, which allows one to mock the remotes behaviour.
+    descr : `str`, optional
+        Short description of the script.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    - "Slewing to ...": Before slewing to target.
+    - "Detection image": If taking in-focus detection image.
+    - "[N/MAX_ITER]: CWFS loop starting...": Before each cwfs iteration, where
+        "N"  is the iteration number and "MAX_ITER" is the maximum number of
+        iterations.
+    - "[N/MAX_ITER]: CWFS converged.": Once CWFS reaches convergence.
+    - "[N/MAX_ITER]: CWFS focus error too large.": If computed focus correction
+        is larger than specified threshhold.
+    - "[N/MAX_ITER]: CWFS applying coma and focus correction.": Just before
+        corrections are applied.
+
+    **Details**
+
+    This script is used to perform measurements of the wavefront error, then
+    propose hexapod offsets based on an input sensitivity matrix to minimize
+    the errors. The hexapod offsets are not applied automatically and must be
+    performed by the user.
+    """
+
+    def __init__(self, index=1, remotes=True, descr=""):
+
+        super().__init__(
+            index=index,
+            descr=descr,
+        )
+
+        self.atcs = None
+        self.latiss = None
+
+        if remotes:
+            self.atcs = ATCS(self.domain, log=self.log)
+            self.latiss = LATISS(
+                self.domain,
+                log=self.log,
+                tcs_ready_to_take_data=self.atcs.ready_to_take_data,
+            )
+
+        # Timeouts used for telescope commands
+        self.timeout_short = 5.0  # used with hexapod offset command
+        self.timeout_long = 30.0  # used to wait for in-position event from hexapod
+        # Have discovered that the occasional image will take 12+ seconds
+        # to ingest
+        self.timeout_get_image = 20.0
+        self.data_pool_sleep = 5.0
+
+        # Sensitivity matrix: mm of hexapod motion for nm of wfs. To figure out
+        # the hexapod correction multiply the calculated zernikes by this.
+        # Note that the zernikes must be derotated before applying the
+        # corrections. See https://tstn-016.lsst.io/ for information about how
+        # these values were derived.
+        self.matrix_sensitivity = [
+            [1.0 / 206.0, 0.0, 0.0],
+            [0.0, -1.0 / 206.0, -(109.0 / 206.0) / 4200],
+            [0.0, 0.0, 1.0 / 4200.0],
+        ]
+
+        # Rotation matrix to take into account angle between camera and
+        # boresight
+        self.matrix_rotation = lambda angle: np.array(
+            [
+                [np.cos(np.radians(angle)), -np.sin(np.radians(angle)), 0.0],
+                [np.sin(np.radians(angle)), np.cos(np.radians(angle)), 0.0],
+                [0.0, 0.0, 1.0],
+            ]
+        )
+
+        # Matrix to map hexapod offset to alt/az offset in the focal plane
+        # units are arcsec/mm. X-axis is Elevation
+        # Measured with data from AT run SUMMIT-5027.
+        self.hexapod_offset_scale = [
+            [52.459, 0.0, 0.0],
+            [0.0, 50.468, 0.0],
+            [0.0, 0.0, 0.0],
+        ]
+
+        # Default values for rotator angle and rotator strategy.
+        self.rot = 0.0
+        self.rot_strategy = RotType.SkyAuto
+
+        # Angle between camera and boresight
+        # Assume perfect mechanical mounting
+        self.camera_rotation_angle = 0.0
+
+        # The following attributes are set via the configuration:
+        self.filter = None
+        self.grating = None
+
+        # exposure time for the intra/extra images (in seconds)
+        self.exposure_time = None
+
+        # Assume the time-on-target is 10 minutes (600 seconds)
+        # for rotator positioning
+        self.time_on_target = 600
+
+        # Set stamp size for WFE estimation
+        # 192 pix is size for dz=1.5, but gets automatically
+        # scaled based on dz later, so can multiply by an
+        # arbitrary factor here to make it larger
+        self._side = 192 * 1.1  # normally 1.1
+
+        # offset for the intra/extra images
+        self._dz = None
+
+        # end of configurable attributes
+
+        # angle between elevation axis and nasmyth2 rotator
+        self.angle = None
+
+        self.intra_visit_id = None
+        self.extra_visit_id = None
+
+        self.zern = None
+        self.hexapod_corr = None
+
+        # Keep track of the total hexapod offset
+        self.offset_total_focus = 0.0
+        self.offset_total_coma_x = 0.0
+        self.offset_total_coma_y = 0.0
+
+    # define the method that sets the hexapod offset to create intra/extra
+    # focal images
+    @property
+    def dz(self) -> float:
+        if self._dz is None:
+            self.dz = 0.8
+        return self._dz
+
+    @dz.setter
+    def dz(self, value: float) -> None:
+        self._dz = float(value)
+        self._run_additional_dz_settings()
+
+    @property
+    def side(self) -> int:
+        # must be an even number
+        return int(np.ceil(self._side * self.dz / 1.5 / 2.0) * 2)
+
+    @abc.abstractmethod
+    async def run_align(self) -> LatissAlignResults:
+        """Runs curvature wavefront sensing code.
+
+        Returns
+        -------
+        results : `LatissAlignResults`
+            A dataclass containing the results of the calculation.
+        """
+        raise NotImplementedError()
+
+    def _run_additional_dz_settings(self) -> None:
+        """Additional actions needed when setting the focus offset value.
+
+        By default do nothing. When subclassing, implement additional actions
+        that need to be executed when setting the focus offset value.
+        """
+        pass
+
+    async def take_intra_extra(self) -> None:
+        """Take pair of Intra/Extra focal images to be used to determine the
+        measured wavefront error.
+
+        Because m2 is being moved, the intra focal image occurs when the
+        hexapod receives a positive offset and is pushed towards the primary
+        mirror. The extra-focal image occurs when the hexapod is pulled back
+        (negative offset) from the best-focus position.
+        """
+
+        self.log.debug("Moving to intra-focal position")
+
+        await self.hexapod_offset(self.dz)
+
+        self.log.debug("Taking intra-focal image")
+
+        intra_image = await self.latiss.take_cwfs(
+            exptime=self.exposure_time,
+            n=1,
+            group_id=self.group_id,
+            filter=self.filter,
+            grating=self.grating,
+            reason="INTRA" + ("" if self.reason is None else f"_{self.reason}"),
+            program=self.program,
+        )
+
+        self.log.debug("Moving to extra-focal position")
+
+        # Hexapod offsets are relative, so need to move 2x the offset
+        # to get from the intra- to the extra-focal position.
+        # then add the offset to compensate for un-equal magnification
+        await self.hexapod_offset(-(self.dz * 2.0 + self.extra_focal_offset))
+
+        self.log.debug("Taking extra-focal image")
+
+        extra_image = await self.latiss.take_cwfs(
+            exptime=self.exposure_time,
+            n=1,
+            group_id=self.group_id,
+            filter=self.filter,
+            grating=self.grating,
+            reason="EXTRA" + ("" if self.reason is None else f"_{self.reason}"),
+            program=self.program,
+        )
+
+        self.intra_visit_id = int(intra_image[0])
+        self.extra_visit_id = int(extra_image[0])
+
+        self.log.info(
+            f"Intra/extra exposure ids: {self.intra_visit_id}/{self.extra_visit_id}"
+        )
+
+        self.angle = 90.0 - await self.atcs.get_bore_sight_angle()
+
+        self.log.info(f"Angle used in cwfs algorithm is {self.angle:0.2f}")
+
+        self.log.debug("Moving hexapod back to zero offset (in-focus) position")
+        # This is performed such that the telescope is left in the
+        # same position it was before running the script
+        await self.hexapod_offset(self.dz)
+
+    async def hexapod_offset(
+        self, offset: float, x: float = 0.0, y: float = 0.0
+    ) -> None:
+        """Applies z-offset to the hexapod to move between
+        intra/extra/in-focus positions.
+
+        Parameters
+        ----------
+        offset : `float`
+             Focus offset to the hexapod in mm.
+        x : `float`, optional
+            Offset the hexapod in the x-axis, in mm (default=0).
+        y : `float`, optional
+            Offset the hexapod in the y-axis, in mm (default=0).
+        """
+
+        offset = {
+            "m1": 0.0,
+            "m2": 0.0,
+            "x": x,
+            "y": y,
+            "z": offset,
+            "u": 0.0,
+            "v": 0.0,
+        }
+
+        self.atcs.rem.ataos.evt_detailedState.flush()
+        await self.atcs.rem.ataos.cmd_offset.set_start(
+            **offset, timeout=self.timeout_long
+        )
+        # Wait for ataos to go through a cycle, which will apply the offset if
+        # enough error has accumulated to pass the ataos thresholds.
+        # Success means we need to see ATAOS substate bit 3 (hexapod
+        # correction) or 4 (Focus correction), then 0 (IDLE).
+        # The bit(s) will flip when the correction is being determined,
+        # regardless  of if a hexapod actually had to move
+
+        start_time = time.time()
+        check_ss2 = False
+        while time.time() < start_time + self.timeout_long:
+            state = await self.atcs.rem.ataos.evt_detailedState.next(
+                flush=False, timeout=self.timeout_long
+            )
+            if (
+                bool(state.substate & (1 << 3))
+                or bool(state.substate & (1 << 4))
+                or check_ss2 is True
+            ):
+                check_ss2 = True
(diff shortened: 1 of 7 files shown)


tribeiro:
Just a note on the test duplication.. I found a way to avoid duplicating it using this library https://pypi.org/project/parameterized/ but it is not part of our stack. :(

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+    zernikes_rot: typing.Tuple[float, float, float]
+    offset_hex: typing.Tuple[float, float, float]
+    offset_tel: typing.Tuple[float, float, float]
+
+
+class LatissBaseAlign(salobj.BaseScript, metaclass=abc.ABCMeta):
+    """Implements generic behavior for scripts that execute curvature wavefront
+    sensing, abstracting the part that performs the measurements.
+
+    This meta Script implements the basic functionality of finding targets and
```
patrickingraham, line 54: what is a "meta script" ?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+    taking the (intra/extra focal) data. Child classes are left to implement
```
patrickingraham, line 55: As a generic comment, should we not be using semantic line versioning for all comments/docstrings etc?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+
+        # Timeouts used for telescope commands
+        self.timeout_short = 5.0  # used with hexapod offset command
+        self.timeout_long = 30.0  # used to wait for in-position event from hexapod
+        # Have discovered that the occasional image will take 12+ seconds
+        # to ingest
+        self.timeout_get_image = 20.0
+        self.data_pool_sleep = 5.0
+
+        # Sensitivity matrix: mm of hexapod motion for nm of wfs. To figure out
```
patrickingraham, line 117: should probably reference the technote here.
https://tstn-016.lsst.io/

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+            [
+                [np.cos(np.radians(angle)), -np.sin(np.radians(angle)), 0.0],
+                [np.sin(np.radians(angle)), np.cos(np.radians(angle)), 0.0],
+                [0.0, 0.0, 1.0],
+            ]
+        )
+
+        # Matrix to map hexapod offset to alt/az offset in the focal plane
+        # units are arcsec/mm. X-axis is Elevation
+        # Measured with data from AT run SUMMIT-5027, still unverified.
```
patrickingraham, line 138: I don't think we have a reason to believe this is incorrect?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+        results : `LatissAlignResults`
+            Results of the wavefront sensing.
+        """
+        rot_zern = np.matmul(
+            self.zern, self.matrix_rotation(self.angle + self.camera_rotation_angle)
+        )
+        hexapod_offset = np.matmul(rot_zern, self.matrix_sensitivity)
+        tel_offset = np.matmul(hexapod_offset, self.hexapod_offset_scale)
+
+        self.log.info(
```
patrickingraham, line 352: I really wish we could publish these results in a formatted way... like write an object to the LFA...
How does the MTAOS do it?
Can we be writing a file here?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+                )
+                if checkpoint:
+                    await self.checkpoint(
+                        f"[{i + 1}/{self.max_iter}]: CWFS focus error too large."
+                    )
+                await self.hexapod_offset(focus_offset / 2.0)
+            else:
+                self.offset_total_focus += focus_offset
+                self.log.info(
+                    f"Applying offset: x={coma_x}, y={coma_y}, z={focus_offset}."
```
patrickingraham, line 755: maybe constrain the values here such as...
`{coma_x:0.3f}`

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+                    same size as the intra. This value is always positive to compensate for
+                    the change in magnification from moving the secondary mirror.
+                type: number
+                default: 0.0011
+              large_defocus:
+                description: >-
+                    Defines a large defocus. If the measured defocus is larger than this value,
+                    apply only half of correction.
+                type: number
+                default: 0.08
```
patrickingraham, line 478: I think we can increase this now... maybe to 0.12?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+              threshold:
+                 description: >-
+                   Focus correction threshold. If correction is lower than this
+                   value, stop correction loop.
+                 type: number
+                 default: 0.01
```
patrickingraham, line 484: Here too, we can boost it to 0.015

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+        can be set to False, which allows one to mock the remotes behaviour.
+    descr : `str`, optional
+        Short description of the script.
+
+    Notes
+    -----
+    **Checkpoints**
+
+    - "Slewing to ...": Before slewing to target.
+    - "Detection image": If taking in-focus detection image.
```
patrickingraham, line 73: what are we detecting? See discussion about this in the config area

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+                default: empty_1
+              exposure_time:
+                description: The exposure time to use when taking intra/extra focal images (sec).
+                type: number
+                default: 30.
+              acq_exposure_time:
+                description: The exposure time to use when taking an in focus acquisition image(sec).
+                type: number
+                default: 5.
+              take_detection_image:
```
patrickingraham, line 456: So this was used for the WEP testing parts before, correct? Is it still required? 
If so, maybe add a sentence in the description saying why it exists and when to use it.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+
+        Parameters
+        ----------
+        metadata : `salobj.type_hints.BaseMsgType`
+            Script metadata topic. The information is set on the topic
+            directly.
+        """
+        # It takes about 300s to run the cwfs code, plus the two acquisition
+        # exposures.
+        metadata.duration = 300.0 + 2.0 * self.exposure_time
```
patrickingraham, line 593: I think we can update this to be max_iterations * exposure_time*2 + detection image + in-focus image time etc. Plus a small fudge factor of course.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+        RuntimeError:
+            If coordinates are malformed.
+        """
+
+        if self.cwfs_target is not None and self.cwfs_target_dec is None:
+            if checkpoint:
+                await self.checkpoint(f"Slewing to object: {self.cwfs_target}")
+            await self.atcs.slew_object(
+                name=self.cwfs_target,
+                rot=0.0,
```
patrickingraham, line 616: Do we need to do this at rot=0 ? It adds overhead, especially when using the parallactic rotType.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+                await self.checkpoint(
+                    f"Slewing to icrs coordinates: {self.cwfs_target} @ "
+                    f"ra/dec = {self.cwfs_target_ra}/{self.cwfs_target_dec}."
+                )
+
+            await self.atcs.slew_icrs(
+                ra=self.cwfs_target_ra,
+                dec=self.cwfs_target_dec,
+                target_name=self.cwfs_target,
+                rot=0.0,
```
patrickingraham, line 636: same deal as above

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -0,0 +1,788 @@
...
+                self.offset_total_coma_x += coma_x
+                self.offset_total_coma_y += coma_y
+                await self.hexapod_offset(focus_offset, x=coma_x, y=coma_y)
+                if self.offset_telescope:
+                    tel_el_offset, tel_az_offset = (
+                        results.offset_tel[0],
+                        results.offset_tel[1],
+                    )
+                    self.log.info(
+                        f"Applying telescope offset az/el: {tel_az_offset}/{tel_el_offset}."
```
patrickingraham, line 770: `{tel_az_offset:0.3f}`

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -204,43 +108,20 @@ def __init__(self, index=1, remotes=True):
...
-        return self._dz
-
-    @property
-    def binning(self):
+    def binning(self) -> int:
         return self._binning
 
     @binning.setter
-    def binning(self, value):
+    def binning(self, value: int) -> None:
```
patrickingraham, line 120: Nice to see this!

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
```diff
@@ -265,128 +146,13 @@ def dz(self, value):
...
-                or check_ss2 is True
-            ):
-                check_ss2 = True
-                if state.substate == 0:
-                    break
-
-    async def run_cwfs(self):
-        """Runs CWFS code on intra/extra focal images.
+    async def run_align(self) -> LatissAlignResults:
+        """Runs curvature wavefront sensing code.
```
patrickingraham, line 150: Maybe we should expand this to emphasize it's the original code and not the WEP code?

Runs CWFS analysis using the ts_cwfs package.

patrickingraham: Somewhere a log info message should also be written that says which analysis is being used.

patrickingraham (changes requested):
First round of comments, most are really minor... but might be worth addressing before I look at the WEP part.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: I guess, I made the term up, But basically, it is a "meta class" that implements a SAL Script hence, a "meta script". In essence, this is not a script that you can use as-is, it needs specializations to implement the core methods that will actually do the wavefront sensing.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: Uhm, not sure I understand what you mean. Can you rephrase the question?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: The MTAOS has events for these, but the results are also stored in the butler by the pipeline task. We can certainly publish them to the FLA tough.

I can make a ticket for that, don't want to make conflate too many changes in one go.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: Yes, I added this for early WEP testing. In my implementation I ended up using the same "ISR" and detection code we use for the old cwfs code so, it is not used for processing the data. I though about leaving it just because it might be a nice debugging tool eventually. But it is right that it adds unnecessary complexity, so maybe I should remove it. 

I'm not sure, what do you think? remove or keep it for debugging purposes and add a note about it?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: We don't need to do it at `rot=0.0` what do you suggest tough? should we add this as a parameter? or maybe try to get the current rotator position and use it in `PhysicalSky` mode?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
tribeiro: I can make this information part of the script description which appears on LOVE at a special/static location.

tribeiro:
@patrickingraham , I think I addressed al your current comments. Some of them I think I need some input to resolve.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
patrickingraham: I think you can leave it, just add a sentence to the description. 
Something like, "Used for early WEP testing but maintained for possible debugging scenarios"

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
patrickingraham: why not make the `rot` and `RotType` config parameters? Default is SkyAuto (so basically do whatever is fastest).

Feeding the current position to PhysicalSky seems like we're adding complication to simplify standard behavior?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
patrickingraham: Ok, that'd be nice. But it goes into the log as well? Just trying to see how one would figure this out when mining back through old data.

patrickingraham (commented):
Sent replies to comments, I'll look at the WEP part shortly...

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: uhm.. I'm not sure that would work for docstrings though... Because we have limits on the line lengths and such.

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
tribeiro: So, instead of going to logMessage, this information would appear in the `Script.logevent_description` topic.. which I think might even make it easier to track down...

image

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
patrickingraham: Not worth holding this up, but shouldn't it be done anyways to the extent possible? Meaning a new sentence goes on a new line? It should reduce the amount of differences when making text changes and even help avoid passing limits.

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
```diff
@@ -0,0 +1,286 @@
...
+        result_extra.brightestObjCentroidCofM[0]
+        - result_intra.brightestObjCentroidCofM[0]
+    )
+    dx = (
+        result_extra.brightestObjCentroidCofM[1]
+        - result_intra.brightestObjCentroidCofM[1]
+    )
+    dr = np.sqrt(dy**2 + dx**2)
+
+    position_out_of_range = dr > 100
```
patrickingraham, line 180: I don't like that this is hardcoded... I know i'm the one that did it... I have mixed feelings about making it a config parameter.

In fact, since the entire catalogue is being used (below), can we just get rid of it?

python/lsst/ts/externalscripts/auxtel/latiss_cwfs_align.py
tribeiro: Not at all..

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
```diff
@@ -0,0 +1,286 @@
...
+    config.donutStampSize = donut_diameter
+    config.donutTemplateSize = donut_diameter
+    config.opticalModel = "onAxis"
+
+    task = EstimateZernikesLatissTask(config=config)
+
+    camera = best_effort_isr.butler.get(
+        "camera",
+        dataId={"instrument": "LATISS"},
+        collections="LATISS/calib/unbounded",
```
patrickingraham, line 201: what is this collection?
I guess i'm a little confused what is going on here. This grabs information about the camera that is needed by the  `EstimateZernikesLatissTask` method?
The reason you don't just put it inside the class is why? Speed? The class is already latiss specific...

tests/auxtel/test_latiss_cwfs_align.py
```diff
@@ -431,32 +430,32 @@ async def test_take_sequence(self):
...
-                f"Hex_tol is {np.matmul(zern_tol, self.script.sensitivity_matrix)}"
+                f"Hex_tol is {np.matmul(zern_tol, self.script.matrix_sensitivity)}"
             )
 
             tel_offset_tol = np.matmul(hex_tol, self.script.hexapod_offset_scale)
             logger.debug(f"tel_offset_tol is {tel_offset_tol}")
 
             results = self.script.calculate_results()
-            for i, z in enumerate(results["zerns"]):
+            for i, z in enumerate(results.zernikes):
```
patrickingraham, line 442: I definitely like the object better than the dictionary.

tests/auxtel/test_latiss_wep_align.py
```diff
@@ -0,0 +1,676 @@
...
+import pytest
+
+try:
+    # TODO: (DM-24904) Remove this try/except clause when WEP is adopted
+    from lsst.ts import wep
+
+    WEP_AVAILABLE = True
+except ImportError:
+    WEP_AVAILABLE = False
+    warnings.warn("Could not import cwfs package. Most tests will be skipped.")
```
patrickingraham, line 38: wep not cwfs

tests/auxtel/test_latiss_wep_align.py
```diff
@@ -0,0 +1,676 @@
...
+# Make obslsst translators less chatty
+logging.getLogger("lsst.obs.lsst.translators").setLevel(logging.WARNING)
+logging.getLogger("astro_metadata_translator").setLevel(logging.WARNING)
+logging.getLogger("flake8.style_guide").setLevel(logging.WARNING)
+
+
+logger = logging.getLogger(__name__)
+logger.propagate = True
+
+random.seed(47)  # for set_random_lsst_dds_partition_prefix
```
patrickingraham, line 59: I don't think this is needed anymore?

tests/auxtel/test_latiss_wep_align.py
```diff
@@ -0,0 +1,676 @@
...
+# Check to see if the test data is accessible for testing at NCSA
```
patrickingraham, line 60: remove "at NCSA"

tests/auxtel/test_latiss_wep_align.py
```diff
@@ -0,0 +1,676 @@
...
+# Depending upon how we load test data this will change
+# for now just checking if butler is instantiated with
+# the default path used on the summit and on the NTS.
+
+# DATAPATH = "/readonly/repo/main" # NTS - obsolete
+DATAPATH = "/repo/LATISS"  # SUMMIT (warning!)
+# DATAPATH = "/repo/LATISS" # TTS (TBC)
```
patrickingraham, line 67: you ran these tests on the TTS right? I think the datapath is the same so you can remove line 67 and make this SUMMIT and TTS

tests/auxtel/test_latiss_wep_align.py
```diff
@@ -0,0 +1,676 @@
...
+
+try:
+    butler = dafButler.Butler(
+        DATAPATH,
+        instrument="LATISS",
+        collections=["LATISS/raw/all", "LATISS_test_data"],
+    )
+    DATA_AVAILABLE = True
+except FileNotFoundError:
+    logger.warning("Data unavailable, certain tests will be skipped")
```
patrickingraham, line 77: why does this use `logger.warnings` whereas line 38 uses `warning.warn`

patrickingraham (commented):
Looks good.
Did you run the tests on the TTS with the data?

tribeiro:
> Looks good.
> Did you run the tests on the TTS with the data?

I did, the wep one is not passing because the results are different. I tried to figure out why and found the problem with the obscuration parameter. But even using the same value for that the results are still different. Not too different, but different enough that is course the checks to fail, given the threshold you set.

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
tribeiro: This is running on a separate process, so it does not have access to the properties of the class. I am not sure what that collection is as well.

tests/auxtel/test_latiss_wep_align.py
tribeiro: I don't know.. I just copied what was in the `test_latiss_cwfs_align.py` file... I think you did that... I actually didn't want to copy and paste it but I didn't find a good way to not duplicate this test... :(

tests/auxtel/test_latiss_wep_align.py
tribeiro: Yes, I ran them on TTS.. All these came from the original file.. I'll cleanup both..

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
```diff
@@ -0,0 +1,286 @@
...
+    ra, dec = wcs.pixelToSkyArray(
+        result.brightestObjCentroidCofM[0],
+        result.brightestObjCentroidCofM[1],
+        degrees=False,
+    )
+
+    donut_catalog = pandas.DataFrame([])
+    donut_catalog["coord_ra"] = ra
+    donut_catalog["coord_dec"] = dec
+    donut_catalog["centroid_x"] = [result.brightestObjCentroidCofM[0]]
```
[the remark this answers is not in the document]
tribeiro: yes.. I'm after doing some tests I decided that it is better to continue using best effort isr and `QuickFrameMeasurementTask` to do pre-process the data and find the donut source. I think we will have to have a conversation about what we want to do with the MT but we have tested best effort isr and quick frame measurement to exhaustion on AuxTel so, I am not ready to gamble on this.

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
tribeiro: well.. not really.. as I mentioned below (sorry I am going through these from the bottom up), I am still using best effort isr and quick frame measurement to pre-process the data and find the donuts so, in a sense, this is still needed. 

I think we can have a conversation about whether we should stop using best effort  isr and quick frame measurement, but we spend quite some time improving these so, I think we should continue using them.

tests/auxtel/test_latiss_wep_align.py
patrickingraham: Does it matter? I honestly didn't know there were two different "types" of warnings...

tribeiro: I think logging and warning have slightly different use-cases.
 
According to the [logging doc page](https://docs.python.org/3/howto/logging.html#when-to-use-logging):

> [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning
>
> [logging.warning()](https://docs.python.org/3/library/logging.html#logging.warning) if there is nothing the client application can do about the situation, but the event should still be noted

I think for this particular one the `warning.warn` is more suitable than the `logging.warning`.

I already modified the code to use `warning.warn`.

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
patrickingraham: Ok. Let's have a chat but we can make changes under a different PR.

patrickingraham (approved):

tribeiro:
I think we need to spend some tie trying to figure out why the results are different in the first place. I tried to figure it out when I was writing the tests but couldn't. One of the reasons I didn't continued with the effort was discovering that the value for the obscuration we're using for cwfs is wrong. I think we need to settle on the correct value before doing additional work.

PR #69, opened 2022-09-15
DM-36153: Add grating to the configurable parameters of make_latiss_calibrations script

plazas:


tribeiro (commented):
Please, update the version history explaining what changes were made. I think this can go on the category of patch so make it version `v0.17.3`...

tribeiro (approved):
Looks good, but please, update the version history before merging.

PR #70, opened 2022-09-21
Tickets/dm 36118

tribeiro:


python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -696,7 +696,35 @@ async def latiss_take_sequence(self, silent=True):
...
+        ataos_corrections = await self.atcs.rem.ataos.evt_correctionEnabled.aget(
+            timeout=self.atcs.fast_timeout
+        )
+
+        assert (
+            ataos_corrections.hexapod
+            and ataos_corrections.m1
+            and ataos_corrections.atspectrograph
+        ), (
+            "Required ATAOS corrections not enabled: ",
```
patrickingraham, line 717: I think this might be a bit ambiguous. Consider:
`Not all required ATAOS corrections are enabled. The following loops must all be closed (True), but are currently:`

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -855,6 +855,31 @@ async def arun(self, checkpoint: bool = False) -> None:
             f"Reached maximum iteration ({self.max_iter}) without convergence.\n"
         )
 
+    async def assert_feasibility(self) -> None:
+        """Verify that the telescope in a feasible state to execute the
```
patrickingraham, line 859: telescope and camera are in....

python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py
```diff
@@ -696,7 +696,35 @@ async def latiss_take_sequence(self, silent=True):
             await self.atcs.rem.ataos.cmd_offset.set_start(z=-self.manual_focus_offset)
             self.manual_focus_offset_applied = False
 
+    async def assert_feasibility(self) -> None:
+        """Verify that the telescope in a feasible state to execute the
```
patrickingraham, line 700: telescope and camera are in a..

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -855,6 +855,31 @@ async def arun(self, checkpoint: bool = False) -> None:
             f"Reached maximum iteration ({self.max_iter}) without convergence.\n"
         )
 
+    async def assert_feasibility(self) -> None:
```
patrickingraham, line 858: You used the same code twice... should it be a utility function? Could be part of a future ticket...

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -855,6 +855,31 @@ async def arun(self, checkpoint: bool = False) -> None:
...
+        ataos_corrections = await self.atcs.rem.ataos.evt_correctionEnabled.aget(
+            timeout=self.atcs.fast_timeout
+        )
+
+        assert (
+            ataos_corrections.hexapod
+            and ataos_corrections.m1
+            and ataos_corrections.atspectrograph
+        ), (
+            "Required ATAOS corrections not enabled: ",
```
patrickingraham, line 876: same coment as above regarding the wording.

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
```diff
@@ -249,7 +249,7 @@ def get_donut_catalog(result: Struct, wcs: SkyWcs) -> pandas.DataFrame:
 
 def get_image(
     data_id: typing.Dict[str, typing.Union[int, str]],
-    best_effort_isr: BestEffortIsr,
+    best_effort_isr: typing.Any,
```
patrickingraham, line 252: .... I honestly don't know what this does...

doc/version_history.rst
```diff
@@ -6,6 +6,67 @@
 Version History
 ===============
 
+v0.17.2
+-------
+
+* In `python/lsst/ts/externalscripts/auxtel/latiss_acquire_and_take_sequence.py`, add feasibility check before executing script.
+  This will check that all CSCs are enabled and that the required ATAOS corrections are enabled.
```
patrickingraham, line 13: all TCS and LATISS controlled CSCs

doc/version_history.rst
```diff
@@ -6,6 +6,67 @@
...
+    * acquire intra/extra focal data by offsetting the hexapod in z,
+    * run a meta function that computes the wavefront errors,
+    * de-rotate the wavefront errors,
+    * apply a sensitivity matrix to compute hexapod and telescope offsets,
+    * apply comma and focus correction to the hexapod and pointing offsets.
+
+  Therefore child implementations are only left to implement the function that computes the wavefront errors.
+
+* In ``LatissCWFSAlign``, use new meta script ``LatissBaseAlign``.
+  This basically remove all the code that was moved from ``LatissCWFSAlign`` into ``LatissBaseAlign``.
```
patrickingraham, line 47: removes

patrickingraham (approved):
Only significant issue is whether or not the assertion method should be made a summit_util...
Please address before merging.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: I honestly don't know. I don't think a utility would be the correct thing to do in this case, because it means we are passing the classes into a method, which is kind of an anti pattern. If I had to swing for the fence here, I would create a Base Script that would have this kind of check in it. Something like having `LatissBaseAlign` inherit `TrackTargetAndTakeImage`.  But yeah, I think we could leave that for a future ticket..

python/lsst/ts/externalscripts/auxtel/latiss_wep_align.py
tribeiro: So, the problem is that `BestEffortIsr` is only available if `summit_utils` is available, and python fails to create the class when it is not. Adding a `typing.Any` is basically telling the type checker (which is external to python) that this object can be of any type.. I'll fix it back in the future once I can incorporate summit_utils into the tssw stack.

PR #71, opened 2022-10-03
Tickets/dm 36314

tribeiro:


b1quint (approved):
I checked the changes. The file with the most updates was the Jenkinsfile. It seems good. I checked the other files, and they seem good as well.

PR #72, opened 2022-10-07
Tickets/sitcom 480

tribeiro:


b1quint (approved):
I reviewed the `building_pointing_model` and its unit tests before the last two commits, and they were fine. The last two commits contained a run of `isort`, which reorganizes the imports on every file. This is why this PR also contains so many changed files. Everything looks fine to me!

PR #73, opened 2022-10-11
Add LOVE stress test script

sebastian-aranda:
This PR adds an external script that simulates client connections in order to stress tests any particular LOVE instance.


tribeiro (changes requested):
Please, use lower case only for the file name, e.g. `make_love_stress_tests.py`...

There are also a lot of things missing to make this a SAL Script able to run on the ScriptQueue... I think it might be better if you issue a ticket for one of us to convert this into a SAL Script... I am not sure it is worth your time to learn all the details involved..

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+            properties:
+              location:
+                description: Domain of the running LOVE instance to stress
+                type: string
+              username:
+                description: Username to authenticate connections
+                type: string
+              password:
+                description: Password to authenticate User
+                type: string
```
r-owen, line 160: username and password will be published to the EFD. I strongly suggest finding a safer way to deal with these secrets, such as using secrets in the Kubernetes deployment to set env vars.

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+import json
+
+import aiohttp
+import requests
+import yaml
+from astropy.time import Time
+from lsst.ts import salobj
+
+
+class ClientInstance:
```
r-owen, line 33: I suggest renaming this to avoid the word "Instance"; that's potentially very confusing to readers because when you instantiate a class the result is an instance (in this case an instance of ClientInstance). Perhaps LoveManagerClient?

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+    """Connect to a LOVE-manager instance.
+
+    Notes
+    -----
+    **Details**
+
+    * Generate websocket connections using provided credentials
+    by token authentication and subscribes to every
```
r-owen, line 41: subscribes -> subscribe

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+
+    async def start_ws_client(self):
+        """Start client websocket connection"""
+
+        async with aiohttp.ClientSession() as session:
+            self.websocket = await session.ws_connect(self.websocket_url)
+            for csc in self.event_streams:
+                for stream in self.event_streams[csc]:
+                    await self.subscribe_to(
+                        csc.split(":")[0], csc.split(":")[1], stream, "event"
```
r-owen, line 110: I gently suggest using `csc_name, sal_index = salobj.name_to_name_index(csc)` and calling this once right after `for csc in self.event_streams`. (or `salindex`; I tend to use snake_case in my code).

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+                    if "category" not in msg or (
+                        "option" in msg and msg["option"] == "subscribe"
+                    ):
+                        continue
+                    self.received_messages = self.received_messages + 1
+                    tracing = msg["tracing"]
+                    tracing["client_rcv"] = Time.now().tai.datetime.timestamp()
+                    self.msg_traces.append(tracing)
+
+    async def subscribe_to(self, csc, salindex, stream, topic_type):
```
r-owen, line 90: This overloads the term `csc`: `start_ws_client` uses `csc` to mean the CSC name + index and here `csc` means just the name. Please use a different variable name here or in `start_ws_client` to clarify the difference.

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+
+    * Run a LOVE stress test by generating several client connections
+    that will listen to every event and telemetry of the specified CSCs
+    """
+
+    def __init__(self, index):
+        super().__init__(index=index, descr="Run a stress test for one or more CSCs")
+
+        # number of clients to simulate on the stress test
+        self.number_of_clients = 10
```
r-owen, line 135: I suggest making this a configuration parameter for the script. It can have a default value, if you like.

Also please consider adding a parameter to control the data rate (e.g. event_interval or something similar).

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+            async for message in self.websocket:
+                if message.type == aiohttp.WSMsgType.TEXT:
+                    msg = json.loads(message.data)
+                    if "category" not in msg or (
+                        "option" in msg and msg["option"] == "subscribe"
+                    ):
+                        continue
+                    self.received_messages = self.received_messages + 1
+                    tracing = msg["tracing"]
+                    tracing["client_rcv"] = Time.now().tai.datetime.timestamp()
```
r-owen, line 87: I strongly recommend `tracing["client_rcv"] = utils.current_tai()`. It is much more efficient than astropy.Time for scalars, and gives the correct form of time (TAI unix seconds) with minimal fuss.

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+        super().__init__(index=index, descr="Run a stress test for one or more CSCs")
+
+        # number of clients to simulate on the stress test
+        self.number_of_clients = 10
+
+        # list to store clients connections
+        self.clients = []
+
+        # commands timeout
+        self.cmd_timeout = 10
```
r-owen, line 141: I suggest adding `self.remotes = dict()` here, since you don't set it until `configure` and that way the attribute is always available. Include a code comment explaining the keys and values.

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+                    event_streams,
+                    telemetry_streams,
+                )
+            )
+
+        for client in self.clients:
+            client.request_token()
+            loop.create_task(client.start_ws_client())
+
+    async def clean(self):
```
r-owen, line 283: When is this intended to be called? I don't see it in the code.

If this is intended to be called at final cleanup then the method name should be `cleanup`. It will be called automatically as the script exits. I also suggest explicitly closing all remotes, though in fact I think the default close method will handle that for you.

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+        """Return the system to its default status."""
+
+        for csc, remote in self.remotes.items():
+            evt = await remote.evt_summaryState.aget()
+            state = salobj.State(evt.summaryState)
+
+            if state == salobj.State.ENABLED:
+                await remote.cmd_disable.start(timeout=self.cmd_timeout)
+                await remote.cmd_standby.start(timeout=self.cmd_timeout)
```
r-owen, line 292: Consider replacing this entire loop body with`await salobj.set_summary_state(remote=remote, state=salobj.State.STANDBY)`. It is simpler and works from any initial state except FAULT.

python/lsst/ts/externalscripts/make_LOVE_stress_tests.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def get_mean_latency(self):
```
r-owen, line 294: Where is this used? I can certainly imagine publishing it as a log message or something.

That's the trickiest thing about using scripts to measure things: there is no dedicated XML Script topic for this purpose. Two choices are to use logMessage or write the data to the lsst `verify` package (which is what test_speed.py does in ts_salobj).

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def __init__(self, index):
+        super().__init__(index=index, descr="Run a stress test for one or more CSCs")
+
+        # number of clients to simulate on the stress test
+        self.number_of_clients = 10
+
+        # number of messages to store
+        # before calculating the latency
+        self.number_of_messages = 1000
```
r-owen, line 137: It'd be nice to know that these are set by configuration (and you should add location, as well). But a simpler, and I think better, solution is to replace these with a `self.config` attribute, initialize it to `None` and then in the `configure` method set it to the provided configuration. Then access these values as `self.config.number_of_clients`, etc.

This simplifies the code, reduces duplication, and eliminates the confusion and hassle of assigning initial values here that never used (since they are overwritten by `configure`).

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+        Specify the Stress test configurations:
+        - LOVE host location
+        - Number of clients
+        - Number of messages to store
+        - CSCs
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Configuration with several attributes:
```
r-owen, line 201: I suggest pointing users to the get_schema method, instead of duplicating that info here (twice -- in this list and in the Notes section below). Duplicated information is hard to keep up to date.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+        """
+        self.log.info("Configure started")
+
+        # set configurations
+        self.location = config.location
+        self.number_of_clients = config.number_of_clients
+        self.number_of_messages = config.number_of_messages
+
+        # get credentials
+        self.username = "user"
```
r-owen, line 234: Any chance this can come from os.environ?

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+        self.password = os.environ.get("USER_USER_PASS")
+
+        if self.password is None:
+            raise Exception("Configuration failed")
```
r-owen, line 238: Please explain what is wrong (and please never raise Exception; always raise a subclass), e.g.:
```
raise RuntimeError("Configuration failed: environment variable USER_USER_PASS not defined")
```

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+    async def run(self):
+        """Run script."""
+        tasks = [
+            remote.start_task
+            for remote in self.remotes.values()
+            if not remote.start_task.done()
+        ]
+        if tasks:
+            self.log.info(f"Waiting for {len(tasks)} remotes to be ready")
+            await asyncio.gather(*tasks)
```
r-owen, line 260: Will you ever have no remotes? What does the script do in that situation? Should it fail?

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+        # before calculating the latency
+        self.number_of_messages = 1000
+
+        # list to store clients connections,
+        # each one an instance of ManagerClient
+        self.clients = []
+
+        # list to store remote connections,
+        # each one an instance of salobj.Remote
+        self.remotes = []
```
r-owen, line 145: This is actually a dict of name_index: Remote. please fix the comment and the initial value.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+            for remote in self.remotes.values()
+            if not remote.start_task.done()
+        ]
+        if tasks:
+            self.log.info(f"Waiting for {len(tasks)} remotes to be ready")
+            await asyncio.gather(*tasks)
+
+        # Enable all CSCs
+        for csc, remote in self.remotes.items():
+            await salobj.set_summary_state(remote=remote, state=salobj.State.ENABLED)
```
r-owen, line 264: You don't use the `csc` variable, so I suggest:

```
for remote in self.remotes.values():
    await...
```

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+
+        event_streams = dict()
+        telemetry_streams = dict()
+        for producer in self.remotes.keys():
+            event_streams[producer] = salobj.SalInfo(
+                self.domain, producer.split(":")[0]
```
r-owen, line 270: Please do NOT create multiple SalInfo. There is no point, and it is expensive. Also this use of getattr is confusing, because you can just get the attribute directly.

Also please use name_to_name_index everywhere. Just look for all instances of `split` in this code.

Consider something like this:
```
for name_index in self.remotes:  # .keys() is optional
    name, index = salobj.name_to_name_index(name_index)
    salinfo = salobj.SalInfo(domain=self.domain, name=name)
    try:
        event_streams[name_index] = salinfo.event_names
        telemetry_streams[name_index] = salinfo.telemetry_names
    finally:
        await salinfo.close()
```
Finally I strongly suggest `name_index` as the variable name for a SAL component name with index. I see at least 3 different variable names in this code: `elt`, `csc`, and `producer`, all of which mean the same thing. `elt` is uninformative, and `csc` and `producer` sound like something much larger (e.g. an instance of salobj.BaseCsc or some class named Producer).

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,318 @@
...
+
+        msg_count = 0
+        while msg_count < self.number_of_messages:
+            await asyncio.sleep(1)
+            new_count = 0
+            for client in self.clients:
+                new_count += len(client.msg_traces)
+            msg_count += new_count
+        mean_latency = round(self.get_mean_latency(), 2)
+        self.log.info(f"Mean latency after {msg_count} messages is: {mean_latency} ms")
```
r-owen, line 304: Please consider tweaking the format to make it easier to parse, e.g.:
```
f"mean_latency_ms={mean_latency} num_messages={msg_count}")
```

r-owen (commented):
Looks much better. Thanks for improving the password handling.

I have some suggestions I hope you will consider.

python/lsst/ts/externalscripts/make_love_stress_tests.py
sebastian-aranda: I think this shouldn't happen as the `config.data` param should always have at least 1 item, defined by the schema:
```
data:
    description: List of CSC_name[:index]
    type: array
    minItems: 1
    items:
```

r-owen: Then why the `if tasks`? I suggest omitting that if statement and always logging and awaiting

sebastian-aranda: I was looking at this and there is case which will have `tasks` as an empty array. As the array is fille with a conditional `if not remote.start_task.done()`, in the case where all remotes start_task param future is done, then the array will be empty.

python/lsst/ts/externalscripts/make_love_stress_tests.py
sebastian-aranda: I think at some point we can define the LOVE default users as environment variables. But at this moment we should have to add those variables for the `admin`, `user`, `cmd_user`, `authlist_user` and `test`. I suggest to leave this fixed for the moment.

python/lsst/ts/externalscripts/make_love_stress_tests.py
r-owen: Fair enough.

It would simpler and more robust to accumulate all of them and await all of them:
```
self.log.info(f"Waiting for {len(self.remotes)} remotes to be ready")
await asyncio.gather(*[remote.start_task for remote in self.remotes.values()])
```
It is more robust because if one of the remote.start_task has already failed, this gather will raise an exception, whereas your current code silently ignores such a failure.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,341 @@
...
+        may take a 10s or 100s of seconds per CSC.
+        """
+        self.log.info("Configure started")
+
+        # set configurations
+        self.location = config.location
+        self.config = {
+            "number_of_clients": config.number_of_clients,
+            "number_of_messages": config.number_of_messages,
+        }
```
r-owen, line 245: I strongly recommend simplifying the lines (all of # set configurations) to:
```
self.config = config
```
Then you can access location as `self.config.location`, number_of_clients as `self.config.number_of_clients`, etc. It avoids repetition and it eliminates the need to use dict style access.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,341 @@
...
+            salinfo = salobj.SalInfo(self.domain, name)
+            try:
+                event_streams[name_index] = salinfo.event_names
+                telemetry_streams[name_index] = salinfo.telemetry_names
+            finally:
+                await salinfo.close()
+
+        # Create clients and listen to ws messages
+        self.log.info(
+            f"Waiting for {self.config['number_of_clients']} Manager Clients to be ready"
```
r-owen, line 296: If you make self.config a copy of config (as suggested above) this will be:
```
f"Waiting for {self.config.number_of_clients} Manager Clients to be ready"
```

r-owen (approved):
Thank you for the cleanups. I have a few more small simplifications that I hope you will consider, but merge when ready.

python/lsst/ts/externalscripts/make_love_stress_tests.py
tribeiro: The password is already coming from an environment variable. I think it would probably be wise to use the same technique for the username. This way we don't have to update the script in the future..

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,324 @@
...
+                event_streams[name_index] = salinfo.event_names
+                telemetry_streams[name_index] = salinfo.telemetry_names
+            finally:
+                await salinfo.close()
+
+        # Create clients and listen to ws messages
+        self.log.info(
+            f"Waiting for {self.config.number_of_clients} Manager Clients to be ready"
+        )
+        loop = asyncio.get_event_loop()
```
tribeiro, line 281: Are you sure you need to do this? Can you use `asyncio.create_task` (in line 294) instead? I think it does the exact same thing.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,324 @@
...
+                    self.config.location,
+                    self.username,
+                    self.password,
+                    event_streams,
+                    telemetry_streams,
+                )
+            )
+
+        for client in self.clients:
+            loop.create_task(client.start_ws_client())
```
tribeiro, line 294: As I mentioned above, I think you can do `asyncio.create_task(...)` instead.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,324 @@
...
+                ManagerClient(
+                    self.config.location,
+                    self.username,
+                    self.password,
+                    event_streams,
+                    telemetry_streams,
+                )
+            )
+
+        for client in self.clients:
```
tribeiro, line 293: Could you do this in the same loop you create the client? maybe something like:

```
            self.clients.append(
                 ManagerClient(
                     self.config.location,
                     self.username,
                     self.password,
                     event_streams,
                     telemetry_streams,
                 )
             )

             asyncio.create_task(self.clients[-1].start_ws_client())
```

?

Just to avoid duplicating `for` loops...

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,324 @@
...
+            loop.create_task(client.start_ws_client())
+
+        msg_count = 0
+        while msg_count < self.config.number_of_messages:
+            await asyncio.sleep(1)
+            new_count = 0
+            for client in self.clients:
+                new_count += len(client.msg_traces)
```
tribeiro, line 301: Do you really need `new_count`? can you do

```
for client in self.clients:
    msg_count += len(client.msg_traces)
```
?

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,324 @@
...
+                    telemetry_streams,
+                )
+            )
+
+        for client in self.clients:
+            loop.create_task(client.start_ws_client())
+
+        msg_count = 0
+        while msg_count < self.config.number_of_messages:
+            await asyncio.sleep(1)
```
tribeiro, line 298: Maybe extract the 1 in the sleep into a constant or a class attribute with a relevant name? Something like `loop_time_message_collection`?

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,324 @@
...
+            new_count = 0
+            for client in self.clients:
+                new_count += len(client.msg_traces)
+            msg_count += new_count
+        mean_latency = round(self.get_mean_latency(), 2)
+        self.log.info(f"mean_latency_ms={mean_latency} num_messages={msg_count}")
```
tribeiro, line 304: Might be beneficial to add some text to facilitate mining this particular data. Something like:


```
self.log.info(f"LOVE stress test result: mean_latency_ms={mean_latency} num_messages={msg_count}")
```

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def set_metadata(self, metadata):
+        """Compute estimated duration.
+
+        Parameters
+        ----------
+        metadata : SAPY_Script.Script_logevent_metadataC
+        """
+        # a crude estimate;
+        metadata.duration = len(self.clients) * 5
```
tribeiro, line 206: I think this will also depend more on `number_of_messages` then on the number of clients no? Maybe create a "expected_message_frequency" parameter to have the multiplying factor? So this would be:

```
metadata.duration = self.config.number_of_messages / self.expected_message_frequency
```

And you define `expected_message_frequency` in the `__init__` method.

tribeiro (commented):
I have a few comments inline for you to consider... Also, can you please rebase your branch to develop? I think there are a couple changes there you need to pickup..

python/lsst/ts/externalscripts/make_love_stress_tests.py
r-owen: I agree. Furthermore, you should basically never use `asyncio.get_event_loop()` as it has unpredictable behavior (it will try to create a loop if none is running, which can cause hard-to-diagnose errors). The preferred alternative is `asyncio.get_running_loop()` when you need a loop, which is rare (my main use is `loop.run_in_executor`)

python/lsst/ts/externalscripts/make_love_stress_tests.py
r-owen: Also this code will not work reliably. You must keep a reference to each task you create until the task finishes. Otherwise you risk it being garbage-collected and canceled before it is done.

Two options are to accumulate the tasks in a separate list, or inside the clients. I usually do the latter (if I wrote the class), because it makes cancelling much easier: the close/cancel/stop method of the object can cancel all such internal tasks. This might do:
```
class Client:
    def __init__(self, ...):
        self.start_task = utils.make_done_future()

    def create_start_task(self):
        self.start_task = asyncio.create_task(self.start_ws_client())
```
The name is a bit ugly, but indicates that the method creates the task without waiting for it to finish. (A name like "start" implies that tit is async and waits until the task finishes).

python/lsst/ts/externalscripts/make_love_stress_tests.py
sebastian-aranda: That's correct. I'll adjust this :+1:

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,333 @@
...
+            additionalProperties: false
+        """
+        return yaml.safe_load(schema_yaml)
+
+    def set_metadata(self, metadata):
+        """Compute estimated duration.
+
+        Parameters
+        ----------
+        metadata : SAPY_Script.Script_logevent_metadata
```
r-owen, line 214: The term `metadata` sounds like data about a topic, rather than message data. I gently suggest `data`.

Also that data type is confusing because SALPY is no longer supported. I suggest something like this (BaseMsgType is especially helpful for type annotations, but it's fine in doc strings as well):
```
data : `lsst.ts.salobj.BaseMsgType`
     Script ``metadata`` event data.
```
All that said: I don't see this method being called. Can you delete it? If I missed where you are calling it, and you are calling it in one place I'd suggest just moving the one line of code there.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,333 @@
...
+
+        msg_count = 0
+        while msg_count < self.config.number_of_messages:
+            await asyncio.sleep(self.loop_time_message_collection)
+            for client in self.clients:
+                msg_count += len(client.msg_traces)
+        mean_latency = round(self.get_mean_latency(), 2)
+        self.log.info(
+            f"LOVE stress test result: mean_latency_ms={mean_latency} num_messages={msg_count}"
+        )
```
r-owen, line 313: I suggest not rounding numeric data (unless necessary) because it discards information. In this case you are just trying to produce nicely formatted output, so specify the desired format in the f string:
```
self.log.info(
    f"LOVE stress test result: mean_latency_ms={self.get_mean_latency():0.2f} num_messages={msg_count}"
)
```

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,333 @@
...
+
+        # Close all ManagerClients
+        for client in self.clients:
+            await client.close()
+
+        # Send all CSCs to Standby
+        for remote in self.remotes.values():
+            await salobj.set_summary_state(remote=remote, state=salobj.State.STANDBY)
+
+        await self.close()
```
r-owen, line 326: I suggest removing this call to `self.close()`, unless you have found it to be necessary (in which case I need to know this, as it's a bug). BaseScript should call `close` when needed.

r-owen (approved):
Looking good. I have some additional minor suggestions I hope you will consider before merging. Sorry to keep dragging this out.

sebastian-aranda:
> Looking good. I have some additional minor suggestions I hope you will consider before merging. Sorry to keep dragging this out.
On the contrary! thanks for the great feedback @r-owen. I'll take a look to the comments now.

python/lsst/ts/externalscripts/make_love_stress_tests.py
sebastian-aranda: I thought this was used by `BaseScript` and one had to override it with the corresponding information. Can you confirm if I'm right? I brought this from https://github.com/lsst-ts/ts_standardscripts/blob/develop/python/lsst/ts/standardscripts/set_summary_state.py#L166

r-owen: You are right. My mistake. Thank you.

python/lsst/ts/externalscripts/make_love_stress_tests.py
```diff
@@ -0,0 +1,364 @@
...
+import yaml
+from lsst.ts import salobj, utils
+
+
+class ManagerClient:
+    """Connect to a LOVE-manager instance.
+
+    Parameters
+    ----------
+    location : `string`
```
r-owen, line 37: Please change ``` `string` ``` to  ``` `str` ``` everywhere (the Python class is ``str``)

python/lsst/ts/externalscripts/make_love_stress_tests.py
r-owen: I still suggest removing this line.

r-owen (approved):
Looks good. Two small requests to consider before merging.

tribeiro (approved):

PR #75, opened 2022-10-14
DM-36422: Have OCPS calibrations script discard N first biases, darks, and flats

plazas:


tribeiro (approved):
Looks good now! thanks!

PR #77, opened 2022-11-03
DM-36331: Update OCPS calib script documentation 

plazas:
Set ``do_gain_from_flat_pair`` to ``True`` by default following DMTN-222, which says that this gain estimate should happen daily.

### DIFF
--- a/python/lsst/ts/externalscripts/base_make_calibrations.py
+++ b/python/lsst/ts/externalscripts/base_make_calibrations.py
@@ -292,7 +292,8 @@ class BaseMakeCalibrations(salobj.BaseScript, metaclass=abc.ABCMeta):
                 default: "-c isr:doDefect=False "
             do_defects:
                 type: boolean
-                descriptor: Should defects be built using darks and flats?
+                descriptor: Should defects be built using darks and flats?. If true, script mode should \
+                    be"BIAS_DARK_FLAT".
                 default: false
             config_options_defects:
                 type: string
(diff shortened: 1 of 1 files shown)


python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -292,7 +292,8 @@ def get_schema(cls):
                 default: "-c isr:doDefect=False "
             do_defects:
                 type: boolean
-                descriptor: Should defects be built using darks and flats?
+                descriptor: Should defects be built using darks and flats?. If true, script mode should \
```
tribeiro, line 295: I think this description is slightly off, isn't it? Consider something like:

> Should defects be built using darks and flats? This configuration will only be in effect if script_mode = BIAS_DARK_FLAT

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -301,7 +302,8 @@ def get_schema(cls):
                 default: "-c isr:doDefect=False "
             do_ptc:
                 type: boolean
-                descriptor: Should a Photon Transfer Curve be constructed from the flats taken?
+                descriptor: Should a Photon Transfer Curve be constructed from the flats taken? If True, the \
```
tribeiro, line 305: Consider rephrasing this a bit to clarify.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -313,8 +315,8 @@ def get_schema(cls):
                 descriptor: Should the gain be estimated from each pair of flats
                     taken at the same exposure time? Runs the cpPtc.yaml# generateGainFromFlatPairs \
                     pipeline. Use the 'config_options_ptc' parameter to pass options to the ISR and \
-                    cpExtract tasks.
-                default: false
+                    cpExtract tasks. If True, the script mode should be "BIAS_DARK_FLAT".
```
tribeiro, line 318: Same as the previous 2 comments..

tribeiro (approved):
Looks good, but please, take a look at the comments about clarifying the descriptions...

PR #81, opened 2023-01-26
Tickets/sitcom 597 - Implement RandomWalk and SerpentWalk

b1quint:
This PR implements the notebook we have used for the TMA Soak Tests as SAL Scripts.

### DIFF
--- a/python/lsst/ts/externalscripts/maintel/random_walk.py
+++ b/python/lsst/ts/externalscripts/maintel/random_walk.py
@@ -25,8 +25,6 @@ import asyncio
 import numpy as np
 import yaml
 from astropy.time import Time
-from lsst.ts import salobj
-from lsst.ts.idl.enums.Script import ScriptState
 from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
 from lsst.ts.observatory.control.utils import RotType
 from lsst.ts.standardscripts.base_track_target import BaseTrackTarget
@@ -63,11 +61,13 @@ class RandomWalk(BaseTrackTarget):
 
     @classmethod
     def get_schema(cls):
-        schema_yaml = """
+        url = "https://github.com/lsst-ts/ts_externalscripts/"
+        path = "python/lsst/ts/externalscripts/maintel/random_walk.py"
+        schema_yaml = f"""
             $schema: http://json-schema.org/draft-07/schema#
-            $id: https://github.com/lsst-ts/ts_externalscripts/python/lsst/ts/externalscripts/maintel/random_walk.py
+            $id: {url}/{path}
             title: RandomWalk v1
-            description: Configuration for running a random walk on sky with slews\ and tracks.
+            description: Configuration for running a random walk on sky with slews and tracks.
             type: object
             properties:
               total_time:
@@ -118,7 +118,7 @@ class RandomWalk(BaseTrackTarget):
                 type: boolean
                 default: False
               ignore:
-                description: >- 
+                description: >-
                   CSCs from the group to ignore in status check. Name must match
                   those in self.group.components, e.g.; hexapod_1.
                 type: array
@@ -227,8 +227,8 @@ class RandomWalk(BaseTrackTarget):
         big_offset_prob=0.1,
         big_offset_radius=9.0,
     ):
-        """Generate Az/El coordinates for a a long time so we can slew and track
-        to these targets.
+        """Generate Az/El coordinates for a a long time so we can slew and
+        track to these targets.
 
         Parameters
         ----------
@@ -252,7 +252,8 @@ class RandomWalk(BaseTrackTarget):
         step = 0
         timer_task = asyncio.create_task(asyncio.sleep(total_time))
         self.log.info(
-            f"{'Time':25s}{'Steps':>10s}{'Old Az':>10s}{'New Az':>10s}{'Old El':>10s}{'New El':>10s}{'Offset':>10s}"
+            f"{'Time':25s}{'Steps':>10s}{'Old Az':>10s}{'New Az':>10s}"
+            f"{'Old El':>10s}{'New El':>10s}{'Offset':>10s}"
         )
 
         n_points = 10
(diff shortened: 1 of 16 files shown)


tribeiro:
I haven't looked too deep into the code yet but just want to drop in one quick general comment.

I think we should probably put these TMA testing script into a sub-module/sub-directory inside ``maintel``, just to separate them from regular use script. Maybe call it ``tma_testing`` or simply ``tma``?

So basically move the code into ``python/lsst/ts/externalscripts/maintel/tma`` and the scripts into ``python/lsst/ts/externalscripts/data/scripts/maintel/tma``?

python/lsst/ts/externalscripts/maintel/random_walk.py
```diff
@@ -0,0 +1,342 @@
...
+import numpy as np
+import yaml
+from astropy.time import Time
+from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
+from lsst.ts.observatory.control.utils import RotType
+from lsst.ts.standardscripts.base_track_target import BaseTrackTarget
+
+
+class RandomWalk(BaseTrackTarget):
+    """Performs offsets of a fixed size in random directions with a probablity
```
tribeiro, line 34: typo:

probablity -> probability

python/lsst/ts/externalscripts/maintel/random_walk.py
```diff
@@ -0,0 +1,342 @@
...
+            min_el=self.config.min_el,
+            max_el=self.config.max_el,
+            big_offset_prob=self.config.big_offset_prob,
+            big_offset_radius=self.config.big_offset_radius,
+        )
+
+        for i, az, el in azel_generator:
+            await self.slew_and_track(az, el, target_name=f"random_walk_{i}")
+
+    def random_walk_azel_by_time(
```
tribeiro, line 219: I really liked how you used the generator here! Great idea!

A couple suggestions, though.

First, consider making this an async generator.

Basically all you need to do is:

```
async def random_walk_azel_by_time(...):
    ...
```

And no other changes are needed in the method itself! Pretty neat!  

But then, you will need to change how your ``run`` method calls it, which brings me to the second suggestion. 

I noticed that you are basically passing ``self.config.something`` as arguments to the method. But all this data is already part of the class, so passing them to the method is kind of an overhead. Instead, consider making ``random_wak_azel_by_time`` have no argument at all. Then, internally, you call directly ``self.config.<parameter>``.

Then, your ``run`` method becomes:

```
         async for i, az, el in self.random_walk_azel_by_time():
             await self.slew_and_track(az, el, target_name=f"random_walk_{i}")
```

Nice and tidy!

python/lsst/ts/externalscripts/maintel/random_walk.py
```diff
@@ -0,0 +1,342 @@
...
+            min_az=self.config.min_az,
+            max_az=self.config.max_az,
+            min_el=self.config.min_el,
+            max_el=self.config.max_el,
+            big_offset_prob=self.config.big_offset_prob,
+            big_offset_radius=self.config.big_offset_radius,
+        )
+
+        for i, az, el in azel_generator:
+            await self.slew_and_track(az, el, target_name=f"random_walk_{i}")
```
tribeiro, line 217: Consider adding a checkpoint here before ``await slew_and_track(...)``. Something like:

```
await self.checkpoint(f"[{i}] Tracking {az=}/{el=}.")
```

python/lsst/ts/externalscripts/maintel/serpent_walk.py
```diff
@@ -0,0 +1,339 @@
...
+            az_grid=self.config.az_grid,
+            el_grid=self.config.el_grid,
+            el_cutoff=self.config.el_cutoff,
+        )
+
+        for i, az, el in azel_generator:
+            self.log.debug(f"Serpent Walk - Target {i} - Moving to {az} / {el}")
+            await self.slew_and_track(az, el, target_name=f"serpent_walk_{i}")
+
+    def azel_grid_by_time(self, total_time, az_grid, el_grid, el_cutoff):
```
tribeiro, line 198: Consider doing the same thing here as I suggested in the other script. Convert this to an async generator with no arguments.

python/lsst/ts/externalscripts/maintel/serpent_walk.py
```diff
@@ -0,0 +1,339 @@
...
+            t = Time.now().to_value("isot")
+            self.log.info(
+                f"{t:25s}{step:10d}{old_az:10.2f}{new_az:10.2f}{old_el:10.2f}{new_el:10.2f}"
+            )
+
+            yield step, new_az, new_el
+            old_az, old_el = new_az, new_el
+            step += 1
+
+    def generate_azel_sequence(self, az_seq, el_seq, el_cutoff=90.0):
```
tribeiro, line 253: So, one thing that I notice in this method is that it does not need to be a class member (e.g. it does not access any `self` value). Consider making this a static method, e.g.:

```
@staticmethod
def generate_azel_sequence(az_seq, el_seq, el_cutoff=90.0):
    ...
```

tests/maintel/test_random_walk.py
```diff
@@ -0,0 +1,84 @@
...
+from lsst.ts import externalscripts, standardscripts
+from lsst.ts.externalscripts.maintel import RandomWalk
+from lsst.ts.idl.enums import Script
+
+
+class TestRandomWalk(
+    standardscripts.BaseScriptTestCase, unittest.IsolatedAsyncioTestCase
+):
+    def setUp(self):
+        self.log = logging.getLogger(__name__)
```
tribeiro, line 33: One thing to keep in mind is that ``setUp`` is called before every test. Maybe use ``setUpClass`` instead? it would look something like this:

```
    @classmethod
    def setUpClass(cls) -> None:
        cls.log = logging.getLogger(__name__)
        cls.log.propagate = True
```

But this is a personal preference, to avoid repeating this for every test.

tests/maintel/test_serpent_walk.py
```diff
@@ -0,0 +1,86 @@
...
+
+from lsst.ts import externalscripts, standardscripts
+from lsst.ts.externalscripts.maintel import SerpentWalk
+from lsst.ts.idl.enums import Script
+
+
+class TestSerpentWalk(
+    standardscripts.BaseScriptTestCase, unittest.IsolatedAsyncioTestCase
+):
+    def setUp(self):
```
tribeiro, line 32: same as above, consider using ``setUpClass``.

tribeiro (commented):
This looks great! I really like the way you used the generators. I have a first round of comments I would like you to consider.

PR #86, opened 2023-02-20
Improve StressLOVE script

sebastian-aranda:
This PR makes some changes after receiving some feedback by a few script executions:

- Add delay to Manager clients creation
- Avoid changing CSCs states, only consult if they are enabled, otherwise an Exception is raised

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -6,6 +6,14 @@
 Version History
 ===============
 
+v0.23.2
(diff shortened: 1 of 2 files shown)


tribeiro:
you need to rebase your branch to develop.. also, please, update the version history

tribeiro (approved):
Looks good!

PR #87, opened 2023-02-21
Tickets/sitcom 668

isotuela:
Hi Tiago, 
Here's the proposal for the script log output in the correct_pointing.py SAL script. This is how it would look now:

Script INFO: porigin_x = -21.178 arcsec, porigin_y = -21.763 arcsec, current_azimuth = 89.477 deg, current_elevation = 60.349 deg

Many thanks! 
Ioana


tribeiro (approved):
Looks great! thanks for the updated!

PR #91, opened 2023-03-02
Quick bug fix in random_walk.py

b1quint:
The ``random_walk_azel_by_time`` function returns four values, but the ``for`` loop expected only three values.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -9,6 +9,14 @@ Version History
 v0.25.0
 -------
 
+* In ``random_walk.py``:
+    * The ``random_walk_azel_by_time`` function now returns a dataclass
+    * Replace ``.get`` calls with ``.aget`` calls 
+    * Fix/improve docstring in RandomWalkData
+    * Remove unused variable ```data```
+    * Remove/improve log messages in ``random_walk_by_time``
+    * Improve random_walk_azel_by_time docstring to explain the name ``origin``
+
 * Add new script ``latiss_acquire.py`` for AuxTel.
   This script is used to slew to a target and center it at a specific position.
   
--- a/python/lsst/ts/externalscripts/maintel/tma/random_walk.py
+++ b/python/lsst/ts/externalscripts/maintel/tma/random_walk.py
@@ -34,10 +34,9 @@ from lsst.ts.standardscripts.base_track_target import BaseTrackTarget
 @dataclass
 class RandomWalkData:
     """Stores information used during the execution of the RandomWalk
-    script. The most critical data are the `new_az` and `new_el`.
-    The script uses them to calculate an offset relative to the initial
-    position on sky of the previous target.
-    This avoids a "trend" in azimuth.
+    script. The script uses `az` and `el` to calculate an offset relative 
+    to the initial position on sky of the previous target. This avoids a 
+    "trend" in azimuth movement.
 
     Attributes
     ----------
@@ -244,10 +243,6 @@ class RandomWalk(BaseTrackTarget):
         """
         counter = 0
         timer_task = asyncio.create_task(asyncio.sleep(self.config.total_time))
-        self.log.info(
-            f"{'Time':25s}{'Steps':>10s}{'Old Az':>10s}{'New Az':>10s}"
-            f"{'Old El':>10s}{'New El':>10s}{'Offset':>10s}"
-        )
 
         n_points = 10
         current_az = np.median(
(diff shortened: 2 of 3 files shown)


tests/maintel/tma/test_random_walk.py
```diff
@@ -121,13 +121,13 @@ def actualPosition(self):
...
-            async for i, az, el, off in self.script.random_walk_azel_by_time():
-                sky_offsets.append(off)
+            async for data in self.script.random_walk_azel_by_time():
+                sky_offsets.append(data.offset)
 
                 # TODO: This should not be necessary
                 # However, if I remove it the loop keeps running forever
                 # It does not seem to happen when running the loop w/ hardware
-                if i == 50:
+                if data.counter == 50:
```
r-owen, line 130: Regarding your TODO: I would fully expect this to run forever unless `random_walk_azel_by_time` has an intrinsic limit in the # of targets it will return. 

If it does, then this tests shows that limit is broken. if it does not, then you need to decide how many targets you want.

Please remove the TODO comment -- after fixing the method, if necessary.

python/lsst/ts/externalscripts/maintel/tma/random_walk.py
```diff
@@ -201,15 +232,17 @@ def set_metadata(self, metadata):
         metadata.duration = self.config.total_time
 
     async def run(self):
-        async for i, az, el in self.random_walk_azel_by_time():
-            await self.checkpoint(f"[{i}] Tracking {az=}/{el=}.")
-            await self.slew_and_track(az, el, target_name=f"random_walk_{i}")
+        async for data in self.random_walk_azel_by_time():
```
r-owen, line 235: What does the "by_time" in the method name mean? This seems to pick a random az/el, but I don't see use of time in that method. Perhaps consider a name such as `get_random_walk_azel_target`?

tests/maintel/tma/test_random_walk.py
```diff
@@ -121,13 +121,13 @@ def actualPosition(self):
             )
 
             sky_offsets = []
-            async for i, az, el, off in self.script.random_walk_azel_by_time():
-                sky_offsets.append(off)
+            async for data in self.script.random_walk_azel_by_time():
```
r-owen, line 124: Way up above in test_run, about line 70 I see you use a mock for random_walk_azel_by_time. That explains why test_run passed even when you broke the script. You also mock slew_and_track, which means this test is testing essentially none of your code.

I suggest mocking the "tcs" instead, if practical, though that will clearly be more work. (I also asked Tiago to consider adding a simulation mode to these classes and copied you on the ticket.)

r-owen (commented):
Looks very good. I am hoping you can beef up the unit test of the run function before final review; let me know what you think.

python/lsst/ts/externalscripts/maintel/tma/random_walk.py
```diff
@@ -262,14 +295,17 @@ async def random_walk_azel_by_time(self):
 
             t = Time.now().to_value("isot")
             self.log.info(
-                f"{t:25s}{step:10d}{current_az:10.2f}{new_az:10.2f}"
+                f"{t:25s}{counter:10d}{current_az:10.2f}{new_az:10.2f}"
```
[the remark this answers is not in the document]
b1quint: This format is useful when running on a Jupyter Notebook because the log messages do not display timestamps. It also displayed the data as a table. However, I agree that we are in a different context here and should use an appropriate format. I applied the suggested changes.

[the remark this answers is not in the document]
r-owen: I very strongly recommend fixing the notebooks to display the date in log messages. If configuring the logger to print to stdout you can specify the format. If printing SAL log messages you can print data.private_sndStamp (after converting to ISO format, if you like -- I usually don't bother because I care more about time differences, which I find easier to compute from floats).

python/lsst/ts/externalscripts/maintel/tma/random_walk.py
b1quint: The "by_time" means this function will generate a random target for a finite time. I will update the docstring to try to be more explicit. Initially, the time was an input parameter, making it easier to see, but Tiago asked to use the configuration directly inside the method.

tests/maintel/tma/test_random_walk.py
b1quint: I **think** I covered what you suggested.

tests/maintel/tma/test_random_walk.py
b1quint: The unit test is running fine now.

tribeiro:
Please, make sure you update the version history. These changes should be part of v0.25.0. I will hold on release until we get this PR and #96 merged.

doc/version_history.rst
```diff
@@ -6,6 +6,17 @@
 Version History
 ===============
 
+v0.25.1
```
tribeiro, line 9: This should be `v0.25.0`

tribeiro (approved):
Looks good.. Please, fix the CI errors before merging.. It is currently failing isort in your Script.

tribeiro (changes requested):
Please, make these changes part of v0.25.0

tribeiro (approved):
Looks good! thanks

PR #93, opened 2023-04-04
Tickets/dm 38260

tribeiro:


python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -328,11 +337,21 @@ async def look_up_table_offsets(
...
 
-        tel_el_offset, tel_az_offset, _ = np.matmul(
-            [x, y, offset], self.hexapod_offset_scale
-        )
+        if self.offset_telescope and (x != 0.0 or y != 0.0 or u != 0.0 or v != 0.0):
+            tel_el_offset_xy, tel_az_offset_xy, _ = np.matmul(
+                [x, y, offset], self.hexapod_offset_scale
+            )
+
+            # Note that Az/El is reversed because u (rotation in x) offsets
```
edennihy, line 345: Curious about this comment, what is actually reversed? Functionally the code looks the same so I can't tell what this refers to.

edennihy (approved):
Code changes look OK, one request to clarify a comment.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: ah! good catch! I wrote that comment down but it ended up being wrong.. so I should remove it..

PR #96, opened 2023-05-03
DM-38437: Create improved RandomWalk SAL Script with Generic Camera

tribeiro:
Opening PR for @b1quint .. it's his branch but it was forgotten. We need to merge this!

### DIFF
--- a/python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
+++ b/python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
@@ -1,0 +1,521 @@
+# This file is part of ts_externalscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["RandomWalkAndTakeImagesGenCam"]
+
+import asyncio
+
+import numpy as np
+import yaml
+
+from lsst.ts import utils
+from lsst.ts.observatory.control import Usages
+from lsst.ts.observatory.control.generic_camera import GenericCamera
+from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
+from lsst.ts.observatory.control.utils import RotType
+from lsst.ts.standardscripts.base_track_target_and_take_image import (
+    BaseTrackTargetAndTakeImage,
+)
+
+from .random_walk import RandomWalk
+
+
+class RandomWalkAndTakeImagesGenCam(BaseTrackTargetAndTakeImage):
+    """Perform random slews while taking images with one or more GenCam.
+
+    Perform offsets of a fixed size in random directions with a probability
+    of performing a large offset also in a random direction. Take images in
+    each position with GenericCamera. Move the Dome whenever the difference
+    between its position and the position of the telescope.
+
+    Parameters
+    ----------
+    index : `int`
+        Index of Script SAL component
+    add_remotes : `bool` (optional)
+        Create remotes to control components (default: `True`)? If False, the
+        script will not work for normal operations. Useful for unit testing.
+    """
+
+    def __init__(self, index, add_remotes: bool = True):
+        super().__init__(
+            index=index,
+            descr="Slew and track random targets on sky with offsets between. "
+            "Each offset might be a small offset or a big offset (both in the configuration). "
+            "After each slew and having the dome in position, it will take images with one or "
+            "more Generic Cameras while tracking. ",
+        )
+        self.config = None
+
+        # Is there a better way to define this?
+        Usages.TakeImageFull = 1 << 4
+
+        self.mtcs_usage, self.gencam_usage = (
+            (None, None) if add_remotes else (MTCSUsages.DryTest, Usages.DryTest)
+        )
+
+        self._mtcs = MTCS(
+            domain=self.domain, log=self.log, intended_usage=self.mtcs_usage
+        )
+        self.gencam_list = None
+
+        # Define timeouts in seconds
+        self.fast_timeout = 1
+        self.slow_timeout = 10
+
+        setattr(
+            RandomWalkAndTakeImagesGenCam,
+            "get_azel_random_walk",
+            RandomWalk.get_azel_random_walk,
+        )
+
+    async def _take_data(self):
+        """Takes data with all the generic cameras"""
+        for i in range(self.config.num_exp):
+            self.log.info(f"Taking exposure #{i+1}")
+            tasks = [
+                asyncio.create_task(
+                    cam.take_object(
+                        exptime=exptime,
+                        group_id=self.group_id,
+                        reason=self.config.reason,
+                        program=self.config.program,
+                    )
+                )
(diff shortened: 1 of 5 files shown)


python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
```diff
@@ -0,0 +1,583 @@
...
+    async def move_dome(self, az: float):
+        """Move the dome to a give position
+
+        Parameters
+        ----------
+        az : float
+            New dome Azimuth position in degrees.
+        """
+        await self.tcs.rem.mtdome.cmd_exitFault.set_start()
+        await asyncio.sleep(5)
```
r-owen, line 385: If you can get rid of this delay, please do so. If not, please add a code comment explaining why it is necessary.

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
```diff
@@ -0,0 +1,583 @@
...
+        await self.tcs.rem.mtdome.cmd_moveAz.set_start(position=az, velocity=0)
+        await self.wait_for_dome_in_position()
+
+    async def random_walk_azel_by_time(self):
+        """Generate Az/El coordinates for a finite amount of time so we can
+        slew and track to these targets.
+        """
+        counter = 0
+        timer_task = asyncio.create_task(asyncio.sleep(self.config.total_time))
```
r-owen, line 394: I strongly suggest keeping the timer in the code that calls this. Then this can be a simple "get" function rather than a generator, which has several advantages:
* It eliminates the need to deal with a counter here -- all it does is return az/el -- which, in turn, lets you get rid of the counter field in the struct.
* By keeping time in the calling code, it becomes simple to abort the last iteration if the time runs out during it. I'm not saying you have to do that, but at least it becomes easy to do so if you decide in the future that you want to.

Also: while using a timer to keep track of when to quit is clever, I am uneasy about it for two reasons:
* It is likely to produce logged warnings if you abort the script, because this timer is still running. That  makes asyncio upset (though it doesn't break anything).
* It is harder to log how much time remains (at least in a way that is fully compatible with the timer) because timers don't make that information available.

A technique that avoids these pitfalls is to record the end time when you start running the script, then check the current time against it:
```
end_tai = utils.current_tai() + self.config.total_time
counter = 0
...
while utils.current_tai() < end_tai:
```
or, if you want to log remaining time (a bit uglier if you want to avoid computing it twice):
```
while True:
    rem_time = end_tai - utils.current_tai()
    if rem_time <= 0:
        return None
    self.log.info(f"Iteration {counter} begins; {rem_time:0.1f} seconds remaining")
```

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
```diff
@@ -0,0 +1,583 @@
...
+
+            # Take data in sync with all the GenericCameras
+            await self.take_data()
+
+    def set_metadata(self, metadata):
+        """Set estimated duration of the script."""
+        metadata.duration = self.config.total_time
+
+    async def slew_and_track(self, az, el, target_name=None):
+        """Slew to and track a new target.
```
r-owen, line 498: Please expand this to explain that the tracking is in RA/Dec and az/el is just the starting point.

This does not appear try to find a star near the specified target, so I am a bit surprised that the generic camera photos reliably show anything useful. (I'm not necessary suggesting searching a catalog -- doing that well to find fields with objects of suitable brightness adds a lot of complexity, but it does make me wonder why this script takes images and if that is safe to do, given the risks of stumbling on a field with a really bright source).

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
```diff
@@ -0,0 +1,583 @@
...
+            azMotion = await self.tcs.rem.mtdome.evt_azMotion.aget()
+        except AttributeError:
+            return
+
+        while not azMotion.inPosition:
+            azMotion = await self.tcs.rem.mtdome.evt_azMotion.aget()
+            await asyncio.sleep(5.0)
+
+        if azMotion.state == 1.0:
+            await self.tcs.rem.mtdome.cmd_exitFault.set_start()
```
r-owen, line 583: I am very concerned about this error handling. If the dome faults then it may have stopped *anywhere*. So why is "in position" valid at this point? At the very least I would expect you to raise an exception or return a flag indicating the problem.

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
```diff
@@ -0,0 +1,583 @@
...
+            az_wrap_strategy=self.config.az_wrap_strategy,
+        )
+
+    async def wait_for_dome_in_position(self):
+        """Wait until the dome is in position"""
+        await asyncio.sleep(20)
+        try:
+            azMotion = await self.tcs.rem.mtdome.evt_azMotion.aget()
+        except AttributeError:
+            return
```
r-owen, line 576: What does the attribute error indicate? What's going on here?

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
```diff
@@ -0,0 +1,583 @@
...
+            ra=self.config.ra,
+            dec=self.config.dec,
+            rot=self.config.rot_sky,
+            rot_type=RotType.Sky,
+            target_name=self.config.name,
+            az_wrap_strategy=self.config.az_wrap_strategy,
+        )
+
+    async def wait_for_dome_in_position(self):
+        """Wait until the dome is in position"""
```
r-owen, line 571: I am very confused and concerned about the error handling in this method, as per comments below. (Though I also suggest dropping this method and moving the logic into move_dome as per a comment below, in which case the error handling questions move to the move_dome method).

Please clearly document what move_dome does if the dome faults.

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
```diff
@@ -0,0 +1,583 @@
...
+        await asyncio.sleep(20)
+        try:
+            azMotion = await self.tcs.rem.mtdome.evt_azMotion.aget()
```
r-owen, line 574: I think this logic has a few issues:
* Race conditions: aget may return stale information -- the dome may not even start to move before you call this method.
* Its unwise to call get repeatedly, yet this method is called repeatedly.

To avoid the race condition and make sure that the dome state is fresh, I strongly suggest the following:
* Delete this method. Move the logic into move_dome.
* In move_dome flush evt_azMotion before commanding motion.
* Ditch the call to aget; just loop over calls to evt_azMotion.next(flush=False) as you do below.

The point of combining this method into move_dome is that users cannot accidentally call it without commanding a dome move (which would make it hang forever if you only call next(flush=False)).

r-owen (commented):
Here are preliminary comments. If practical, please digest these and then I'll have another look.

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
b1quint: It does not always have stars in the field, and we do not care much. Maybe I will replace "target" with "field" or "position on sky." The field of these cameras are usually very large, so we usually have a couple of stars.

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
b1quint: Not really sure, I'm removing this.

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
b1quint: I don't remember exactly why I added this. So I will remove it for now.

python/lsst/ts/externalscripts/maintel/tma/random_walk_and_take_image_gencam.py
b1quint: I followed your recommendation and moved the code in `wait_for_dome_in_position` to `move_dome`

b1quint:
Closing this PR to start a new one fresh.

PR #98, opened 2023-05-04
Enabled using HD_cwfs_stars in auxtel/latiss_base_align.py and auxtel…

edennihy:
…/correct_pointing.py.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -11,8 +11,8 @@ v0.25.0
 
 * Add new script ``latiss_acquire.py`` for AuxTel.
   This script is used to slew to a target and center it at a specific position.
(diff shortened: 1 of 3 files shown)


doc/version_history.rst
```diff
@@ -11,6 +11,8 @@ v0.25.0
 
 * Add new script ``latiss_acquire.py`` for AuxTel.
   This script is used to slew to a target and center it at a specific position.
+* In ``auxtel/correct_pointing`` and ``auxtel/latiss_base_align``, add config to search local catalog and set
+  to HD_cwfs_stars by default.
```
tribeiro, line 15: This should be in the line above. We use semantic line breaking in docs so each phrase should be in a single line.

tribeiro (approved):
Looks good! Before merging please, make sure you fix the line in the version history!

PR #100, opened 2023-05-15
Tickets/dm 37831

dsanmartim:
Hi Tiago, Patrick

I've added a few lines of code to return the hexapod to its original position in case of failures (if it does not converge or gets an exception). I'm unsure if this is the best way to do it, so please let me know if anything is wrong or if I'm missing something.

Best,
David.

### DIFF
--- a/python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
+++ b/python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
@@ -22,6 +22,7 @@
 __all__ = ["LatissBaseAlign"]
 
 import abc
+import asyncio
 import dataclasses
 import types
 import typing
@@ -30,6 +31,7 @@ import numpy as np
 import yaml
 from lsst.ts import salobj
 from lsst.ts.idl.enums.ATPtg import WrapStrategy
+from lsst.ts.idl.enums.Script import ScriptState
 from lsst.ts.observatory.control.auxtel.atcs import ATCS
 from lsst.ts.observatory.control.auxtel.latiss import LATISS
 from lsst.ts.observatory.control.constants import atcs_constants
@@ -181,6 +183,9 @@ class LatissBaseAlign(salobj.BaseScript, metaclass=abc.ABCMeta):
 
         self.camera_playlist = None
 
+        # Flag to monitor if iterations have started for cleanup task.
+        self.iterations_started = False
+
     # define the method that sets the hexapod offset to create intra/extra
     # focal images
     @property
@@ -230,6 +235,9 @@ class LatissBaseAlign(salobj.BaseScript, metaclass=abc.ABCMeta):
 
         self.log.debug("Moving to intra-focal position")
 
+        # Updated total focus offset value for intra-focal position
+        self.offset_total_focus += self.dz
+
         await self.atcs.offset_aos_lut(z=self.dz)
 
         self.log.debug("Taking intra-focal image")
@@ -250,6 +258,10 @@ class LatissBaseAlign(salobj.BaseScript, metaclass=abc.ABCMeta):
         # to get from the intra- to the extra-focal position.
         # then add the offset to compensate for un-equal magnification
         z_offset = -(self.dz * 2.0 + self.extra_focal_offset)
+
+        # Updated total focus offset value for extra-focal position
+        self.offset_total_focus += z_offset
+
         await self.atcs.offset_aos_lut(z=z_offset)
 
         self.log.debug("Taking extra-focal image")
@@ -461,7 +473,7 @@ Telescope offsets [arcsec]: {(len(tel_offset) * '{:0.1f}, ').format(*tel_offset)
                    Focus correction threshold. If correction is lower than this
                    value, stop correction loop.
                  type: number
-                 default: 0.008
+                 default: 0.004
               coma_threshold:
                  description: >-
                    Coma correction threshold. If correction is lower than this
@@ -741,25 +753,23 @@ Telescope offsets [arcsec]: {(len(tel_offset) * '{:0.1f}, ').format(*tel_offset)
         self.offset_total_focus = 0.0
         self.offset_total_coma_x = 0.0
         self.offset_total_coma_y = 0.0
-        for i in range(self.max_iter):
-            self.log.debug(f"CWFS iteration {i + 1} starting...")
 
+        # Flagging that iterations have started
+        self.iterations_started = True
+        for self.iterations_executed in range(self.max_iter):
+            self.log.debug(f"CWFS iteration {self.iterations_executed + 1} starting...")
             if checkpoint:
                 await self.checkpoint(
-                    f"[{i + 1}/{self.max_iter}]: CWFS loop starting..."
+                    f"[{self.iterations_executed + 1}/{self.max_iter}]: CWFS loop starting..."
                 )
-
             # Setting visit_id's to none so run_cwfs will take a new dataset.
             self.intra_visit_id = None
             self.extra_visit_id = None
             await self.take_intra_extra()
-
             results = await self.run_align()
-
             coma_x = results.offset_hex[0]
             coma_y = results.offset_hex[1]
             focus_offset = results.offset_hex[2]
-
             total_coma_offset = np.sqrt(coma_x**2.0 + coma_y**2.0)
             if (
                 abs(focus_offset) < self.threshold
@@ -774,12 +784,13 @@ Telescope offsets [arcsec]: {(len(tel_offset) * '{:0.1f}, ').format(*tel_offset)
                     f"Total coma-y correction: {self.offset_total_coma_y:0.3f} mm."
                 )
                 if checkpoint:
-                    await self.checkpoint(f"[{i + 1}/{self.max_iter}]: CWFS converged.")
+                    await self.checkpoint(
+                        f"[{self.iterations_executed + 1}/{self.max_iter}]: CWFS converged."
+                    )
                 # Add coma offsets from previous run
                 self.offset_total_coma_x += coma_x
                 self.offset_total_coma_y += coma_y
                 await self.atcs.offset_aos_lut(z=focus_offset, x=coma_x, y=coma_y)
-
                 current_target = await self.atcs.rem.atptg.evt_currentTarget.aget(
                     timeout=self.timeout_short
                 )
@@ -801,7 +812,6 @@ Telescope offsets [arcsec]: {(len(tel_offset) * '{:0.1f}, ').format(*tel_offset)
                     program=self.program,
                 )
                 await self.atcs.add_point_data()
-
                 self.log.info("latiss_cwfs_align script completed successfully!\n")
                 return
             elif abs(focus_offset) > self.large_defocus:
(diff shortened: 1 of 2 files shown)


dsanmartim:
Damm it... I was trying to make a rebase of my branch after asking for the PR and for some reason I have incorporated all changes from the develop branch as if they were part of my branch. I am really sorry for that. Is there a way to revert this? Maybe deleting my branch and starting it over again? Or maybe with a `git reset --hard origin/tickets/DM-37831`?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -840,97 +840,133 @@ async def arun(self, checkpoint: bool = False) -> None:
...
-
-            # Setting visit_id's to none so run_cwfs will take a new dataset.
-            self.intra_visit_id = None
-            self.extra_visit_id = None
-            await self.take_intra_extra()
-
-            results = await self.run_align()
+        # Getting initial hexapod position
+        self.initial_hexapod_position = await self.latiss.get_hexapod_position()
+        try:
```
tribeiro, line 846: I think that rather than doing a try/except here, you could handle this in the `cleanup` method. This is a coroutine that is automatically called when the script is terminating. 

Take a look at how the cleanup is used in the [base_track_target_and_take_image](https://github.com/lsst-ts/ts_standardscripts/blob/05ad34eee66b7e7b3f29bf6ea5b301cdd9ce0e36/python/lsst/ts/standardscripts/base_track_target_and_take_image.py#L290) script.

tribeiro (commented):
I have one general comment about the implementation. Please, take a look and let me know what you think.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
tribeiro: Also, check the documentation on the cleanup method [here](https://ts-salobj.lsst.io/sal_scripts.html#cleanup-method-optional).

dsanmartim:
Need to reopen the pull request, please.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -829,13 +836,37 @@ async def arun(self, checkpoint: bool = False) -> None:
                 self.offset_total_coma_y += coma_y
                 await self.atcs.offset_aos_lut(z=focus_offset, x=coma_x, y=coma_y)
 
+            # Not sure if here is the best place to get number of executed
```
tribeiro, line 839: So, there's already a iterator being used to track down number of iterations. I think you can either do `self.executed_iterations = i` at the start of the loop, or maybe even switch from using `i` as the iterator to using `self.executed_iterations`, e.g.:

```
for self.executed_iterations in range(self.max_iter):
    ...
```

I think this one is a bit more elegant actually although, you would have to replace all references to `i`. 

I would also recommend renaming the variable to something like `iterations_executed`, it is always best practice to start a variable name with a noun.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -741,25 +746,29 @@ async def arun(self, checkpoint: bool = False) -> None:
         self.offset_total_focus = 0.0
         self.offset_total_coma_x = 0.0
         self.offset_total_coma_y = 0.0
+
+        # Getting initial hexapod position
+        self.initial_hexapod_position = await self.latiss.get_hexapod_position()
```
tribeiro, line 751: The `LATISS` class deals only with the camera, it doesn't have this method defined in it. This would most likely be a method in the `ATCS` class, which also doesn't have such a method. 

I think you can either create it as part of this ticket, or implement something here that will get it for you. Either way works fine, however, this might be a good addition to the `ATCS` class so if you prefer to implement it here, we should probably have a ticket to port it to ts_observatory_control.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -829,13 +836,37 @@ async def arun(self, checkpoint: bool = False) -> None:
...
         self.log.warning(
-            f"Reached maximum iteration ({self.max_iter}) without convergence.\n"
+            f"Reached maximum iteration ({self.max_iter}) without convergence."
+            f"Returning hexapod to initial position. \n"
+        )
+
+        self.rem.ataos.evt_detailedState.flush()
+
+        # Return hexapod to original position
+        await self.atcs.rem.ataos.cmd_offset.set_start(
```
tribeiro, line 852: I am not sure this is going to do what you expect. The `cmd_offset` command is the same command used in `atcs.offset_aos_lut` and what it does is to apply offsets to the hexapod, not set the position. 

So I think this might be missing some steps where it gets the current position of the hexapod and undoes the offsets.

Also, I see you "reset" the hexapod position in 2 different locations, maybe extract this into a `reset_hexapod` method? I would recommend using the `atcs.offset_aos_lut` method though, instead of the command directly.

tribeiro (changes requested):
I left a few inline comments for your consideration. 

Also, make sure you update the version history file.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
dsanmartim: Perfect.  I changed the iterator variable (and all appearances) to `self. iterations_executed`.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
dsanmartim: I got it. There are already variables called `offset_total_focus/coma_x/coma_y` that could be used to return the hexapod back to its original position by applying the total offsets in the reverse direction, like this:

```
await self.atcs.offset_aos_lut(z=-offset_total_focus, x=-offset_total_coma_x, y=-offset_total_coma_y)
```

Does it make sense for you?

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
dsanmartim: Lesson learned: don't trust autocompletion. What a silly mistake :(  

I think that using the `atcs.offset_aos_lut` method in the way I'm proposing below would eliminate the need to have a
`atcs.get_hexapod_position()` method here. In any case, it would be useful to implement it on another ticket. 

Please, let me know your thoughts on this.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
dsanmartim: I implemented that using the `offset_total_focus/coma_x/coma_y` variables as entry to the `atcs.offset_aos_lut`. This may return the hexapod to the initial position, removing the total computed offsets.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -880,3 +894,37 @@ async def run(self) -> None:
...
+                    f"Terminating with state={self.state.state}: stop iterations. "
+                    f"Iterations started: {self.iterations_started}. "
+                )
+                try:
+                    self.log.warning(
+                        f"Looping failed at iter {self.iterations_executed} from a total of "
+                        f"{self.max_iter}. Returning hexapod to its initial position.\n"
+                    )
+
+                    self.rem.ataos.evt_detailedState.flush()
```
tribeiro, line 912: I think here you will have to do something else. Because the loop could be interrupted while it was taking the intra/extra focal image. I think in this situation what you want to do is get the offsets before the script started executing, then get the offsets at this point and apply the offset to restore them to the initial position.

I think you could just put this procedure in a method and then call it here.

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -823,19 +826,30 @@ async def arun(self, checkpoint: bool = False) -> None:
...
         self.log.warning(
-            f"Reached maximum iteration ({self.max_iter}) without convergence.\n"
+            f"Reached maximum iteration ({self.max_iter}) without convergence."
+            f"Returning hexapod to initial position. \n"
+        )
+
+        self.rem.ataos.evt_detailedState.flush()
+
+        # Return hexapod to original position
+        await self.atcs.offset_aos_lut(
```
tribeiro, line 844: I wonder if this is actually what we want to do. In some cases it doesn't converge but it is more aligned than before, in some other cases it might be a divergence of some type. I think most times it is the first case and we are good to keep the offsets. Maybe @edennihy could comment.

tribeiro (commented):
Here some additional comments..

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
dsanmartim: Good point. If it reaches this point in the loop, it means the new values are bigger than the threshold (coma and focus). That is why I added it here, but maybe we need to add one more condition here to check the real need to return it to its initial position. Or just remove it? Cc @edennihy

edennihy: I think the safest case for now is to return to original position. We should be able to control the frequency of the first case by adjusting the threshold values. Recently, after tightening the focus value threshold we have been seeing more cases where the focus is improving, but the coma is diverging.

edennihy: I think the additional logic we would need to capture that should be a separate ticket after we finish our experiments with the threshold values.

edennihy: @dsanmartim, @tribeiro and I talked through this again at lunch today and decided that the original code behavior is best. If after reaching the iteration limit we still do not converge, we should leave the hexapod in it's current position after the corrections from the final iteration have been applied. 

To put it another way, we do **not** want to return to the original position in the case where we do not converge. Sorry for the confusion!

python/lsst/ts/externalscripts/auxtel/latiss_base_align.py
```diff
@@ -880,3 +887,37 @@ async def run(self) -> None:
...
+                    self.log.warning(
+                        f"Loop failed at iteration {self.iterations_executed} from a total of "
+                        f"{self.max_iter}. Returning hexapod to its initial position."
+                    )
+
+                    self.rem.ataos.evt_detailedState.flush()
+
+                    # Return hexapod to original position
+                    await self.atcs.offset_aos_lut(
+                        z=-self.offset_total_focus,
```
tribeiro, line 909: I think this could work if you keep track of the focus offset you apply when taking the intra/extra focal data. 

Basically, before you do [this](https://github.com/lsst-ts/ts_externalscripts/blob/82671cf7a246e1eb74c3a88dbb665dd78ba709cf/python/lsst/ts/externalscripts/auxtel/latiss_base_align.py#L233) and [this](https://github.com/lsst-ts/ts_externalscripts/blob/82671cf7a246e1eb74c3a88dbb665dd78ba709cf/python/lsst/ts/externalscripts/auxtel/latiss_base_align.py#L253) you would do:

```
self.offset_total_focus += self.dz
```

and 

```
self.offset_total_focus += z_offset
```

respectively.

tribeiro (approved):
Alright! looks good!

PR #104, opened 2023-06-12
LOVE-194: Add Uptime test for LOVE

sebastian-aranda:
This PR adds the `make_love_uptime_tests.py` script in order to have an approach to measure the uptime of the system.


r-owen (changes requested):
Looks promising. I have a lot of suggestions, but most are minor.

My main worries are changing log levels for CSCs and not having a time limit for reading summaryState from each CSC (it seems likely the script will often hang without those).

In addition to my suggestion below, I gently suggest making the polling interval a config parameter with a default value (so users will not need to set it, but can -- and so that they can see what value it will be). What you have is fine, but it'd be a nice touch.

tribeiro (commented):
Code looks good but please, organize the commits and fix the version history.

tribeiro (approved):
Looks good! thanks!

PR #105, opened 2023-06-13
DM-39631: SAL script sends NaN offsets when no source is found

psferguson:
Added check that will raise RuntimeError if offsets are NaN, before sending to atcs.


python/lsst/ts/externalscripts/auxtel/correct_pointing.py
```diff
@@ -325,6 +325,12 @@ async def _center(self) -> float:
         )
 
         offset_x, offset_y = await self.find_offset(image_id=acquisition_image_ids[0])
+
+        if np.isnan(offset_x) | np.isnan(offset_y):
```
tribeiro, line 329: I think to compare the two booleans returned by `np.isnan` it is best to use the `or` operator instead of the bitwise `|` comparison. Even if they yield the same results, at the very least, using `or` is more readable.

tests/auxtel/test_correct_pointing.py
```diff
@@ -87,6 +88,18 @@ async def test_configure(self):
 
                 self.assert_config(default_values, config)
 
+    async def test_check_center(self):
+        self.remotes_needed = False
+        async with self.make_script():
+            self.script.find_offset = unittest.mock.AsyncMock(
+                return_value=(np.nan, np.nan)
+            )
```
tribeiro, line 96: Consider mocking `self.script.atcs.offset_xy` then asserting that it is not awaited afterwards?

Basically do:

```
self.script.atcs.offset_xy = unites.mock.AsyncMock()
```

then call 

```
self.script.atcs.offset_xy.assert_not_awaited()
```
after the 

```
with self.assertRaises(RuntimeError):
    await self.script._center()
````

tribeiro (commented):
This looks really good! I only have 2 inline comments I hope you will consider before merging. One in in the code itself (I think it is better to use the `or` operator instead of the bitwise comparison `|`) and the other one in the unit test, to assert that the offset method was not called.

psferguson:
Made the changes you suggested, confirmed the unit tests still pass

PR #118, opened 2023-11-07
DM-41536: Refactor LOVE stress and uptime tests for configuring URLs instead of domains

sebastian-aranda:
PR to make tests more flexible so they can be used also for k8s deployments for instance. Also some docstrings were standardized and `make_love_stress_tests.py` was refactored to import the LOVEManagerClient class instead of defining it.


rbovill (approved):
LGtM

rbovill (changes requested):
Could you add a set of unit tests to verify the secure location, in addition to the non-secure URL?

rbovill (approved):
Thanks for adding the secure location test.  LGtM.

PR #123, opened 2024-01-31
DM-41684: Implement the group_id subgrouping feature for intra/extra pair

psferguson:
A supplemental group id is created before images are taken and this is used for both the intra and extra focal image


tribeiro (approved):
Looks good!

PR #129, opened 2024-05-31
DM-44231: Support Summit Observing Weeks 19-22 of 2024

tribeiro:


edennihy (approved):
Looks good, and all tested on-sky during the week of CR1.

PR #137, opened 2024-07-25
Tickets/dm 45406

dsanmartim:
Hi Tiago. Here is a draft pull request for your review.

### DIFF
--- a/python/lsst/ts/externalscripts/base_take_ptc_flats.py
+++ b/python/lsst/ts/externalscripts/base_take_ptc_flats.py
@@ -1,0 +1,443 @@
+# This file is part of ts_externalscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["BaseTakePTCFlats"]
+
+import abc
+import asyncio
+import types
+
+import yaml
+from lsst.ts import salobj
+from lsst.ts.standardscripts.base_block_script import BaseBlockScript
+from lsst.ts.xml.enums.Electrometer import UnitToRead
+
+
+class BaseTakePTCFlats(BaseBlockScript, metaclass=abc.ABCMeta):
+    """Base class for taking PTC flats interleaved with darks."""
+
+    def __init__(
+        self, index, descr="Base script for taking PTC flats and darks."
+    ) -> None:
+        super().__init__(index, descr)
+
+        self.electrometer = None
+        self.config = None
+
+        self.instrument_setup_time = 0.0
+        self.extra_scan_time = 1.0
+
+        self.long_timeout = 30
+
+    @property
+    @abc.abstractmethod
+    def camera(self):
+        raise NotImplementedError()
+
+    async def configure_camera(self):
+        """Abstract method to configure the camera, to be implemented
+        in subclasses.
+        """
+        raise NotImplementedError()
+
+    async def configure_electrometer(self, index):
+        """Configure the Electrometer remote object."""
+        if self.electrometer is None:
+            self.log.debug(f"Configuring remote for Electrometer index: {index}")
+            self.electrometer = salobj.Remote(
+                self.domain, name="Electrometer", index=index
+            )
+            await self.electrometer.start_task
+        else:
+            self.log.debug("Electrometer already configured. Ignoring.")
+
+    @classmethod
+    def get_schema(cls):
+        schema_yaml = """
+            $schema: http://json-schema.org/draft-07/schema#
+            $id: https://github.com/lsst-ts/ts_externalscripts/base_take_ptc_flats.yaml
+            title: BaseTakePTCFlats
+            description: Configuration schema for BaseTakePTCFlats.
+            type: object
+            properties:
+              flats_exp_times:
+                description: >
+                  A list of exposure times for the flats (sec). Each provided
+                  time will be used to take two exposures sequentially.
+                type: array
+                items:
+                  type: number
+                  exclusiveMinimum: 0
+                minItems: 2
+              interleave_darks:
+                description: Darks interleave settings.
+                type: object
+                properties:
+                  dark_exp_times:
+                    description: >
+                      Exposure time for each dark image (sec)
+                      or alist of exposure times.
+                    anyOf:
+                      - type: number
+                        minimum: 0
+                      - type: array
+                        items:
+                          type: number
+                          minimum: 0
+                        minItems: 1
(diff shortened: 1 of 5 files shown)


python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+                else:
+                    self.log.warning(
+                        f"Component {comp} not in CSC Group. "
+                        f"Must be one of {self.camera.components_attr}. "
+                        f"Ignoring."
+                    )
+
+        # Ensure flats_exp_times is provided in pairs
+        if len(config.flats_exp_times) % 2 != 0:
+            raise ValueError("Flats exposure times must be provided in pairs.")
```
tribeiro, line 183: Should be `RuntimeError`.

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+
+        # Check that each pair in flats_exp_times contains equal values
+        for i in range(0, len(config.flats_exp_times), 2):
+            if config.flats_exp_times[i] != config.flats_exp_times[i + 1]:
+                raise ValueError(
```
tribeiro, line 188: Should be `RuntimeError`

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+                        f"Component {comp} not in CSC Group. "
+                        f"Must be one of {self.camera.components_attr}. "
+                        f"Ignoring."
+                    )
+
+        # Ensure flats_exp_times is provided in pairs
+        if len(config.flats_exp_times) % 2 != 0:
+            raise ValueError("Flats exposure times must be provided in pairs.")
+
+        # Check that each pair in flats_exp_times contains equal values
```
tribeiro, line 185: Why are you enforcing it this way? Why not let the user provide a single value and take twice the number of exposures?

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+                    "Flats exposure times must be provided in pairs of equal values."
+                )
+
+        # Handle interleave darks settings
+        if hasattr(config, "interleave_darks"):
+            if (
+                isinstance(config.interleave_darks["dark_exp_times"], list)
+                and config.interleave_darks["n_darks"]
+            ):
+                raise ValueError(
```
tribeiro, line 198: Should be `RuntimeError`. Also, why not just ignore this altogether? just log a warning saying that `n_darks` are going to be ignored because `dark_exp_times` is an array.

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+        mode : `str`
+            Electrometer measurement mode.
+        range : `float`
+            Electrometer measurement range. -1 for autorange.
+        integration_time : `float`
+            Electrometer integration time.
+        """
+        try:
+            electrometer_mode = getattr(UnitToRead, mode).value
+        except AttributeError:
```
tribeiro, line 261: Why convert an `AttributeError` into a `ValueError`. If anything (which I don't think is necessary) convert it into a `RuntimeError`... But I would just not this at all. Just do 

```
electrometer_mode = getattr(UnitToRead, mode).value
```

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+            raise ValueError(f"Invalid electrometer mode: {mode}")
+
```
tribeiro, line 263: What if the electrometer is not defined? Above you only call `configure_electrometer` if `electrometer_scan` is passed to the configuration. 

I would add an `assert self.electrometer is not None` here. But also, if you always need an electrometer make sure you make `electrometer_scan` required. 

```
        if hasattr(config, "electrometer_scan"):
            await self.configure_electrometer(config.electrometer_scan["index"])
```

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+                        f"Taking flat {j+1}/2 with exposure time: {exp_time} "
+                        f"seconds and scanning electrometer for "
+                        f"{exp_time + self.extra_scan_time} seconds."
+                    )
+
+                    electrometer_task = self.take_electrometer_scan(
+                        exp_time + self.extra_scan_time
+                    )
+
+                    flat_task = self.camera.take_imgtype(
```
tribeiro, line 397: Why not use `take_flats`? you know this is going to be a flat.

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+                    )
+
+                    await asyncio.gather(electrometer_task, flat_task)
+
+                else:
+                    self.log.info(
+                        f"Taking flat {j+1}/2 with exposure time: {exp_time} "
+                        f"seconds."
+                    )
+                    await self.camera.take_imgtype(
```
tribeiro, line 414: Again, use `take_flats`

python/lsst/ts/externalscripts/base_take_ptc_flats.py
```diff
@@ -0,0 +1,451 @@
...
+
+                if hasattr(self.config, "interleave_darks"):
+                    for k, dark_exp_time in enumerate(
+                        self.config.interleave_darks["dark_exp_times"]
+                    ):
+                        self.log.info(
+                            f"Taking dark {k+1}/{len(self.config.interleave_darks['dark_exp_times'])} "
+                            f"for pair {i // 2 + 1} of {len(self.config.flats_exp_times) // 2}."
+                        )
+                        await self.camera.take_imgtype(
```
tribeiro, line 432: Use `take_darks`

tests/maintel/test_take_ptc_flats_comcam.py
```diff
@@ -0,0 +1,106 @@
...
+
+            assert self.script.config.flats_exp_times == [7.25, 7.25, 0.75, 0.75]
+            assert self.script.config.electrometer_scan["index"] == 201
+            assert self.script.config.electrometer_scan["mode"] == "CURRENT"
+            assert self.script.config.electrometer_scan["range"] == -1
+            assert self.script.config.electrometer_scan["integration_time"] == 0.1
+            assert self.script.config.interleave_darks["n_darks"] == 2
+            assert self.script.config.interleave_darks["dark_exp_times"] == [30, 30]
+
+    async def test_invalid_config(self):
```
tribeiro, line 64: Maybe give a hint of what is invalid about it.

tribeiro (commented):
I have a few inline comments.

python/lsst/ts/externalscripts/base_take_ptc_flats.py
dsanmartim: Ok. But the idea is to make it optional.

dsanmartim: And the code you suggested is already done in the he `configure` method:

```
      if hasattr(config, "electrometer_scan"):
            await self.configure_electrometer(config.electrometer_scan["index"])
```

tests/maintel/test_take_ptc_flats_comcam.py
dsanmartim: I was testing an invalid electrometer mode but actually, this will be actually an enum in the schema.

tribeiro (approved):
Looks good to me..

PR #143, opened 2024-08-13
tickets/DM-41364: Fix doc page

cvillalon:
Fixing doc page that was showing ``ts_standardscripts`` instead of ``ts_externalscripts``.


tribeiro (approved):
Looks good

PR #144, opened 2024-08-14
Add parameter_march scripts

gmegh:


python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,442 @@
...
+                description: Azimuth position to point to.
+                type: number
+                minimum: 0
+                maximum: 360
+              el:
+                description: Elevation position to point to.
+                type: number
+                minimum: 0
+                maximum: 90
+              filter:
```
jmeyers314, line 116: I'm wondering on the merits of including `filter` as a parameter here, as opposed to simply always using the current filter (and relegating a filter change to a separate script as necessary).  Changing a filter is a fairly big deal for LSSTCam (though less so for ComCam), so it feels like it maybe shouldn't be buried in a separate script; it should be its own script.  I don't think the same applies to any of the other parameters here (the initial slew could also be a big deal, but since the script ratchets in between triplets I don't really see an alternative).  Curious what @tribeiro and others think.

gmegh: I am not sure if there has been a resolution for this, but FWIW I think the ability to change the filter like this is already present in other scripts. We can just leave it blank if we don't want to change it?

doc/news/DM-45761.feature.rst
```diff
@@ -0,0 +1,4 @@
+- - Introduced a suite of scripts for taking sensitivity matrices and parameter marches with LSSTCam and LSSTComCam:
```
tribeiro, line 1: Instead of adding multiple entries in one file, create one individual file for each entry. For each additional file after the first one, append a number to the file name, e.g.;

- doc/news/DM-45761.feature.rst
- doc/news/DM-45761.feature.rst.1
- doc/news/DM-45761.feature.rst.2

python/lsst/ts/externalscripts/base_parameter_march.py
tribeiro: You are right that changing the filter for LSSTCam is big deal. I have to think about this, for now, I would suggest we leave it here.

python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,461 @@
...
+                    self.log.debug(f"Ignoring Camera component {comp}.")
+                    setattr(self.camera.check, comp, False)
+                else:
+                    self.log.warning(
+                        f"Component {comp} not in CSC Groups. "
+                        f"Must be one of {self.tcs.components_attr} or "
+                        f"{self.camera.components_attr}. Ignoring."
+                    )
+
+        if hasattr(config, "step_sequence"):
```
tribeiro, line 249: Not sure I understand what you are doing here. Why are you modifying things in the `config` object? Why not make whatever parameter you need to extract here a class attribute instead?

python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,461 @@
...
+            config.range = config.step_sequence[-1] - config.step_sequence[0]
+            config.n_steps = len(config.step_sequence)
+        elif hasattr(config, "range"):
+            config.step_sequence = np.linspace(
+                -config.range, config.range, config.n_steps
+            ).tolist()
+
+        if hasattr(config, "rotation_sequence"):
```
tribeiro, line 257: same comment as above. I would rather you leave `config` object alone and create class attributes to hold what you need here.

python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,461 @@
...
+            if isinstance(config.rotation_sequence, (int, float)):
+                config.rotation_sequence = [config.rotation_sequence] * config.n_steps
+            elif isinstance(config.rotation_sequence, list):
+                if len(config.rotation_sequence) != config.n_steps:
+                    raise ValueError(
+                        f"rotation_sequence length {len(config.rotation_sequence)} "
+                        f"does not match n_steps {config.n_steps}."
+                    )
+            else:
+                raise TypeError("rotation_sequence must be either a number or a list.")
```
tribeiro, line 267: Isn't this enforced by the schema already?

python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,461 @@
...
+
+    async def assert_feasibility(self) -> None:
+        """Verify that the telescope and camera are in a feasible state to
+        execute the script.
+        """
+        await asyncio.gather(
+            self.tcs.assert_all_enabled(), self.camera.assert_all_enabled()
+        )
+
+    async def format_values(self, offset_values: np.ndarray) -> np.ndarray:
```
tribeiro, line 296: This method is not using any attribute from the class. Maybe make it a `@staticmethod`?

python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,461 @@
...
+
+        Returns
+        -------
+        str
+            Instrument filter configuration.
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def take_images(self) -> None:
```
tribeiro, line 433: Can't you make `take_images` be the same for both cameras using `self.camera`?

python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,461 @@
...
+
+        self.config = None
+        self.dofs = np.zeros(50)
+
+        self.total_offset = 0.0
+        self.iterations_started = False
+
+    @property
+    @abc.abstractmethod
+    def tcs(self):
```
tribeiro, line 70: The `tcs` is the same for both cameras, no need to make it an abstract property.

python/lsst/ts/externalscripts/base_parameter_march.py
```diff
@@ -0,0 +1,461 @@
...
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    async def configure_tcs(self):
```
tribeiro, line 74: This should be the same for both cameras as well.

python/lsst/ts/externalscripts/maintel/parameter_march_comcam.py
```diff
@@ -0,0 +1,217 @@
...
+
+        self.mtcs = None
+        self.comcam = None
+
+        self.dz = 1500  # microns, offset for out-of-focus images
+
+        self.instrument_name = "LSSTComCam"
+
+    @property
+    def tcs(self):
```
tribeiro, line 57: See how this is just the same method for ComCam and LSSTCam? It should be in the base class.

python/lsst/ts/externalscripts/maintel/parameter_march_comcam.py
```diff
@@ -0,0 +1,217 @@
...
+                self.config.exp_time
+                + self.camera.read_out_time
+                + self.camera.shutter_time
+            )
+        )
+
+        metadata.instrument = self.get_instrument_name()
+        metadata.filter = self.get_instrument_filter()
+
+    async def take_images(
```
tribeiro, line 86: you should be able to implement this method in the base class.

tribeiro (changes requested):
Here is an initial round of comments. I stopped while reviewing the comcam script as I think there are changes there that will impact the rest of the code.

python/lsst/ts/externalscripts/base_parameter_march.py
gmegh: well, for ComCam take images are triplets, while for LSSTCam take images is only taking on image. That's why I had separated. Additionally, in the case of ComCam we ping RA to start processing the data

python/lsst/ts/externalscripts/maintel/parameter_march_comcam.py
gmegh: See comment above. Why would I not separate it, since ComCam will be taking triplets while LSSTCam takes only one image? Doesn't it make it cleaner as is?

doc/news/DM-45761.feature.rst
gmegh: Although, it looks like the Towncrier tests are complaining now...

python/lsst/ts/externalscripts/base_parameter_march.py
tribeiro: you need to remove the `self` argument for a static method.

python/lsst/ts/externalscripts/base_parameter_march.py
tribeiro: Is LSSTCam really taking only one image? I think the idea here would be to do intra/extra regardless no?

gmegh: Well, LSSTCam gets intra focal and extra focal images directly from the corners for free, so we don't need to move the focal plane, unless we are using the "full array mode"

tribeiro (approved):

PR #152, opened 2024-10-23
Tickets/dm 46969 Create script for take_rotated_comcam.py

iglesu:
https://rubinobs.atlassian.net/browse/DM-46969


python/lsst/ts/externalscripts/maintel/take_rotated_comcam.py
```diff
@@ -0,0 +1,186 @@
+# This file is part of ts_standardscripts
```
tribeiro, line 1: ts_externalscripts

python/lsst/ts/externalscripts/maintel/take_rotated_comcam.py
```diff
@@ -0,0 +1,186 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["TakeRotatedComCam", "Mode"]
```
tribeiro, line 22: I don't think you need to re-export `Mode`.

python/lsst/ts/externalscripts/maintel/take_rotated_comcam.py
```diff
@@ -0,0 +1,186 @@
...
+        """Configure script.
+
+        Parameters
+        ----------
+        config : `types.SimpleNamespace`
+            Script configuration, as defined by `schema`.
+        """
+        await super().configure(config)
+        self.config = config
+        self.angles = config.angles
```
tribeiro, line 139: you are already storing the entire `config` object, I don't think you need to also store `angles` as a separate attribute.

python/lsst/ts/externalscripts/maintel/take_rotated_comcam.py
```diff
@@ -0,0 +1,186 @@
...
+            await self.mtcs.slew_icrs(
+                ra=self.config.ra,
+                dec=self.config.dec,
+                rot=angle,
+                rot_type=RotType.Physical,
+                target_name=self.config.target_name,
+                az_wrap_strategy=WrapStrategy.NOUNWRAP,
+            )
+
+            await self.mtcs.wait_for_inposition(
```
tribeiro, line 168: you don't need to do this here.. `slew_icrs` will only finish when it is in position.

tribeiro (commented):
This looks really good! I only have a few comments..

python/lsst/ts/externalscripts/maintel/take_rotated_comcam.py
iglesu: good catch!

python/lsst/ts/externalscripts/maintel/take_rotated_comcam.py
```diff
@@ -0,0 +1,179 @@
...
+__all__ = ["TakeRotatedComCam"]
+
+import astropy.units
+import yaml
+from astropy.coordinates import ICRS, Angle
+from lsst.ts.observatory.control.utils import RotType
+from lsst.ts.standardscripts.maintel.take_aos_sequence_comcam import (
+    TakeAOSSequenceComCam,
+)
+from lsst.ts.xml.enums.ATPtg import WrapStrategy
```
tribeiro, line 31: This should be `MTPtg` not `ATPtg`

tribeiro (approved):
looks good, but I found one more issue with the imports. see my inline comment.

python/lsst/ts/externalscripts/maintel/take_rotated_comcam.py
iglesu: DARN!!! Of course! good catch!!

PR #158, opened 2024-11-07
tickets/DM-47291 Update BaseMakeCalibrations to trigger cp_verify and don't wait for it to finish.

iglesu:
https://rubinobs.atlassian.net/browse/DM-47291

- Refactor run_block to handle calibration and verification concurrently using asyncio
- Added helper methods (process_images, process_verification, process_calibration) to reduce code duplication
- Manage background tasks with a list, including timeout handling and cancellation if not completed in time
- Add configuration option `background_task_timeout` to control background task timeouts
- Added unit test for BaseMakeCalibrations

### DIFF
--- a/python/lsst/ts/externalscripts/base_make_calibrations.py
+++ b/python/lsst/ts/externalscripts/base_make_calibrations.py
@@ -17,6 +17,7 @@
 # GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 __all__ = ["BaseMakeCalibrations"]
 
@@ -71,6 +72,8 @@ class BaseMakeCalibrations(BaseBlockScript, metaclass=abc.ABCMeta):
             "DEFECTS",
             "PTC",
         ]
+
+        self.background_tasks = []
 
         # Pipetask methods to get parameters for calibrations generation
         self.pipetask_parameters = dict(
@@ -347,6 +350,10 @@ class BaseMakeCalibrations(BaseBlockScript, metaclass=abc.ABCMeta):
                 type: integer
                 default: 120
                 descriptor: Timeout value, in seconds, for OODS.
+            background_task_timeout:
+                type: integer
+                default: 30
+                descriptor: Timeout value, in seconds, for background tasks
             note:
                 description: A descriptive note about the images being taken.
                 type: string
@@ -1403,6 +1410,7 @@ class BaseMakeCalibrations(BaseBlockScript, metaclass=abc.ABCMeta):
         process = await asyncio.create_subprocess_shell(cmd)
         stdout, stderr = await process.communicate()
         self.log.debug(f"Process returned: {process.returncode}")
+
         if process.returncode != 0:
             self.log.debug(stdout)
             self.log.error(stderr)
@@ -1547,6 +1555,7 @@ class BaseMakeCalibrations(BaseBlockScript, metaclass=abc.ABCMeta):
         # Basic sets of calibrations first : biases, darks, and flats.
         # After the loop is done, do defects and PTC.
         for im_type in image_types:
+
             # 1. Take images with the instrument, only for "BIAS,
             # "DARK", or "FLAT".
             if im_type == "BIAS":
@@ -1557,8 +1566,8 @@ class BaseMakeCalibrations(BaseBlockScript, metaclass=abc.ABCMeta):
                 await self.checkpoint(f"Taking {self.config.n_flat} flats.")
 
             # TODO: Before taking flats with LATISS (and also
-            # with LSSTComCam), check that the telescope is in
-            # position to do so. See DM-31496, DM-31497.
+            #  with LSSTComCam), check that the telescope is in
+            #  position to do so. See DM-31496, DM-31497.
             exposure_ids_list = await self.take_images(im_type)
 
             # Discard the first N exposures taken (DM-36422)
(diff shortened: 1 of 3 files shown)


python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1547,6 +1555,9 @@ async def run_block(self):
         # Basic sets of calibrations first : biases, darks, and flats.
         # After the loop is done, do defects and PTC.
         for im_type in image_types:
+            if self.certify_calib_failed:
+                break
```
tribeiro, line 1559: we should keep going even if certification fails.

tests/test_base_make_calibrations.py
```diff
@@ -1,3 +1,23 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
```
tribeiro, line 20: This is missing one line here at the end:

```
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
```

I know some other files are missing this line as well. But let's make that fix on a different PR.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1683,26 +1603,138 @@ async def run_block(self):
...
+            if self.config.generate_calibrations:
+                self.log.info(
+                    "'do_verify' is set to 'False' and "
+                    "'generate_calibrations' to 'True'. "
+                    f"{im_type} will be automatically certified."
+                )
+
+                await self.certify_calib(im_type, job_id_calib)
+
+    async def process_verification(self, im_type, job_id_calib):
```
tribeiro, line 1652: please add docstring do the method.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1683,26 +1603,138 @@ async def run_block(self):
...
+            # (bias->dark->flat).
+            else:
+                self.certify_calib_failed = True
+
+            return report_check_verify_stats
+            # Note: Since we are not generating calibrations, we don't certify
+        except Exception as e:
+            self.log.exception(f"Error in processing verification for {im_type}: {e}")
+
+    async def process_calibration(self, calib_type):
```
tribeiro, line 1691: Please add docstring.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1683,26 +1603,138 @@ async def run_block(self):
...
 
-                    self.log.info(f"{calib_type} generation job ID: {job_id_calib}")
+        await self.wait_for_background_tasks()
 
-                    # Report the estimated gain from each pair of flats
-                    if calib_type in ["GAIN", "PTC"]:
-                        await self.report_gains_from_flat_pairs(job_id_calib)
-                except Exception:
-                    self.log.exception(f"Error processing {calib_type}. Ignoring...")
+    async def process_images(self, im_type):
```
tribeiro, line 1613: Please, add docstring.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1683,26 +1603,138 @@ async def run_block(self):
...
+
+        if self.config.generate_calibrations:
+            # 2. Call the calibration pipetask via the OCPS
```
tribeiro, line 1616: Is there a step 1? You might want to remove the comment here. The log message below expains what it is doing.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1683,26 +1603,138 @@ async def run_block(self):
...
+        else:
+            self.log.info(
+                f"A combined {im_type} will not be generated from the "
+                "images taken as part of this script. Any needed input "
+                "calibrations by the verification pipetasks will be "
+                "sought in their input calibrations."
+            )
+            job_id_calib = None
+
+        # 3. Verify the combined calibration (implemented so far for bias,
```
tribeiro, line 1633: Please, remove this comment here. It is sort of obvious what it is doing. If you need to make it more obvious wrap this in a method.

python/lsst/ts/externalscripts/base_make_calibrations.py
```diff
@@ -1683,26 +1603,138 @@ async def run_block(self):
...
+        # dark, and flat), and certify it if the verification
+        # tests pass and it was generated.
+        if self.config.do_verify:
+            try:
+                await self.process_verification(im_type, job_id_calib)
+            except Exception:
+                self.log.exception("Error in do_verify. Ignoring...")
+        # do verify is False
```
tribeiro, line 1641: also obvious what is happening, no need for a comment.

tribeiro (commented):
Here are a few inline comments.

There is one operational issue (we should keep going with calibrations even if certification failed. The others are more style guide, e.g., add docstrings to the methods you created.

Also, I see that there are some comments from the original code that are not super helpful, and only state things that are pretty obvious. I would suggest taking a look at the comments and removing the ones that are redundant.

python/lsst/ts/externalscripts/base_make_calibrations.py
iglesu: This is a change from current behavior, should I create a news fragment for that?

python/lsst/ts/externalscripts/base_make_calibrations.py
iglesu: I don't think so, these are comments I left from the original.

tribeiro (approved):
Looks good.

PR #161, opened 2024-11-18
Tickets/dm 47381: Support Summit Observing Weeks 45-46 of 2024

edennihy:


python/lsst/ts/externalscripts/base_take_twilight_flats.py
```diff
@@ -282,6 +295,9 @@ async def configure(self, config: types.SimpleNamespace):
                 if comp in self.camera.components_attr:
                     self.log.debug(f"Ignoring Camera component {comp}.")
                     setattr(self.camera.check, comp, False)
+                elif comp in ["mtdome", "mtdometrajectory"]:
```
tribeiro, line 298: why make it only `mtdome` and `mtdometrajectory`? shouldn't this be 

```
elif comp in self.tcs.components_attr:
```

?

python/lsst/ts/externalscripts/base_take_twilight_flats.py
```diff
@@ -355,9 +371,26 @@ def get_target_radec(self):
...
 
+        min_sun_distance = 60
+
         az_sun, el_sun = self.tcs.get_sun_azel()
 
-        target_az = (az_sun + self.config.distance_from_sun) % 360
+        if self.config.point_directly:
+            if np.abs(az_sun - (self.config.target_az % 360)) < min_sun_distance:
+                raise RuntimeError(
+                    f"Distance from sun {az_sun - (self.config.target_az % 360)} is \
```
tribeiro, line 381: Please, split the string like we normally do:

```
f"Distance from sun {az_sun - (self.config.target_az % 360)} is "
f"less than {min_sun_distance}. Stopping."
""
```

python/lsst/ts/externalscripts/base_take_twilight_flats.py
```diff
@@ -355,9 +371,26 @@ def get_target_radec(self):
...
+                        less than {min_sun_distance}. Stopping."
+                )
+
+            target_az = self.config.target_az
+        else:
+            if np.abs(self.config.distance_from_sun) < min_sun_distance:
+                raise RuntimeError(
+                    f"Distance from sun {self.config.distance_from_sun} is less than {min_sun_distance}. \
```
tribeiro, line 389: Same as above, split the string like we normally do:

```
f"Distance from sun {self.config.distance_from_sun} is less than {min_sun_distance}. "
"Stopping."
```

python/lsst/ts/externalscripts/base_take_twilight_flats.py
```diff
@@ -513,11 +548,20 @@ async def take_twilight_flats(self):
...
+                    sleep_time = 30
+                    self.log.warning(
+                        f"Calculated exposure time {exp_time} below min exposure time \
+                            {self.config.min_exp_time}. Waiting {sleep_time}s."
+                    )
+                    await asyncio.sleep(sleep_time)
+                    continue
+                else:
+                    self.log.warning(
+                        f"Calculated exposure time {exp_time} below min exposure time \
```
tribeiro, line 561: Please, split the string like we normally do.

python/lsst/ts/externalscripts/data/scripts/maintel/make_cbp_throughput_scan.py
tribeiro, line 1: This should not be included as part of this PR.

python/lsst/ts/externalscripts/maintel/make_cbp_throughput_scan.py
tribeiro, line 1: Please do not include new scripts as part of the run branch. The author needs to open a separate PR to include this.

python/lsst/ts/externalscripts/maintel/parameter_march_comcam.py
```diff
@@ -118,6 +119,14 @@ async def take_images(
             reason="EXTRA" + ("" if self.reason is None else f"_{self.reason}"),
             program=self.config.program,
         )
+        self.log.info("Waiting for data to be ingested by OODS.")
+        for _ in range(9):
```
tribeiro, line 123: I think I did this one but we should not have a hard-coded number here. Can you make this a class attribute?

tests/maintel/test_make_cbp_throughput_scan.py
tribeiro, line 1: This should not be part of this PR.

doc/news/DM-47381.feature.1.rst
tribeiro, line 1: This should not be part of this PR.

python/lsst/ts/externalscripts/base_take_twilight_flats.py
edennihy: I think the idea here was that we only wanted to allow the option to ignore the mtdome and mtdometrajectory componets. For example, if a user ignored the mtmount, mtdome, and mtdometrajectory, we were worried it would leave open the option for the telescope to be pointed in the direction of the sun while taking data.

tribeiro (approved):

PR #164, opened 2024-11-26
Tickets/DM-47791

kfanning:
make_lsstcam_calibrations.py

### DIFF
--- a/python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
+++ b/python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
@@ -1,0 +1,198 @@
+# This file is part of ts_externalscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["MakeLSSTCamCalibrations"]
+
+import yaml
+from lsst.ts.observatory.control import RemoteGroup
+from lsst.ts.observatory.control.maintel.lsstcam import LSSTCam
+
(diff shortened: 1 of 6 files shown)


python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+
+    def __init__(self, index=1):
+        super().__init__(
+            index=index,
+            descr="Takes series of bias, darks and flat-field exposures"
+            "with LSSTCam, and constructs combined "
+            "calibrations, verify and certify the results.",
+        )
+        self._lsstcam = LSSTCam(domain=self.domain, log=self.log)
```
tribeiro, line 48: When you rebase your branch do develop you will notice that now the `BaseMakeCalibrations` script has a new abstract method called `start_remote`. This method is awaited during the configure method and is a better place to initialize the remote. You should do something like:

```
self._lsstcam = None
self._ocps_group = None
```

Here and then do something like:

```
async def start_remotes(self):
    if self._lsstcam is None:
        self._lsstcam = LSSTCam(domain=self.domain, log=self.log)
        await self._lsstcam.start_task

    if self._ocps_group is None:
        self._ocps_group = RemoteGroup(
            domain=self.domain, components=["OCPS:2"], log=self.log
        )
        await self._ocps_group.start_task
```

tests/maintel/test_make_lsstcam_calibrations.py
```diff
@@ -0,0 +1,120 @@
...
+logger.propagate = True
+
+
+class TestMakeLSSTCamCalibrations(
+    standardscripts.BaseScriptTestCase, unittest.IsolatedAsyncioTestCase
+):
+    async def basic_make_script(self, index):
+        logger.debug("Starting basic_make_script")
+        self.script = MakeLSSTCamCalibrations(index=index)
+
```
tribeiro, line 39: To prevent the Script from creating the camera and ocsp objects, make sure you add:

```
self.script._lsstcam = AsyncMock()
self.script._ocps_group = AsyncMock()
```

here. You will have to `from unittest.mock import AsyncMock` at the beginning.

tribeiro (commented):
This looks good but I have some inline comments I hope you will take into account before we can merge this. Also, we have this request to be able to add a delay between exposures, I think we could do that in this ticket and my proposal would be to update the base script, so this can be applied to all the instruments. 

Basically, my suggestion would be to add a `wait_between_exposures` configuration parameter with default value of zero, and then add a 

```
await asyncio.sleep(self.config.wait_between_exposures)
```

somewhere around [here](https://github.com/lsst-ts/ts_externalscripts/blob/302a4ce86c06fd44f412e552a64838d9081be951/python/lsst/ts/externalscripts/base_make_calibrations.py#L479-L494).

That return statement would have to be refactored to add the sleep, but I think it is doable.

tribeiro (approved):

PR #172, opened 2025-01-20
Improve the warmup_hexapod.py.

teweitsai:
Improve the warmup_hexapod.py to recover from the failure and change the step size in the runtime.

### DIFF
--- a/python/lsst/ts/externalscripts/maintel/warmup_hexapod.py
+++ b/python/lsst/ts/externalscripts/maintel/warmup_hexapod.py
@@ -23,10 +23,9 @@ __all__ = ["WarmUpHexapod"]
 import asyncio
 from collections.abc import Iterable
 
-import numpy as np
 import yaml
 from lsst.ts import salobj
-from lsst.ts.idl.enums.MTHexapod import SalIndex
+from lsst.ts.idl.enums.MTHexapod import EnabledSubstate, SalIndex
 from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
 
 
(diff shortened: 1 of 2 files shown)


python/lsst/ts/externalscripts/maintel/warmup_hexapod.py
```diff
@@ -213,31 +212,133 @@ async def warm_up(self):
...
+                if position_next <= stop:
+                    position_next = stop
+
+            # Do the movement
+            all_positions = await self._get_position()
+            all_positions[self.config.axis] = position_next
+
+            is_successful = await self._move_hexapod(**all_positions)
+
+            # Update the step size based on the result
```
b1quint, line 253: I wonder if we should add some verbosity here regarding the changes in the step size. Especially because, if I understand correctly, the step size will be updated on every step. It will either be larger or smaller. But never the same. Is this correct?

teweitsai: Do you mean to add more comment to explain the update process of step size? Or you want to me to add `self.log.info()` or `self.log.debug()` to log the new step size?

teweitsai: This should be resolved in the commit: [0349e62](https://github.com/lsst-ts/ts_externalscripts/pull/172/commits/0349e6263737f4acad2f602f66a29bfb99c9b8a8). Thanks!

b1quint (approved):
We tested this code on TTS and at the summit. We had interactions to debug and confirm that the log messages would give enough information to the night crew. We also confirmed that the script will recover the CSC that faults during warm up. Everything is working great. Thank you Te-Wei!

PR #179, opened 2025-03-17
DM-49346: Park Calibration Projector

parfa30:
After using the Calibration projector, for either the white light or monochromatic flats, the projector should be moved so that it is in a safe place and all LEDs turned off. This script accomplishes that.

### DIFF
--- a/python/lsst/ts/externalscripts/maintel/park_calibration_projector.py
+++ b/python/lsst/ts/externalscripts/maintel/park_calibration_projector.py
@@ -1,0 +1,107 @@
+# This file is part of ts_externalscripts
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["ParkCalibrationProjector"]
+
+import yaml
+from lsst.ts import salobj
+from lsst.ts.observatory.control.maintel.mtcalsys import MTCalsys
+
+
+class ParkCalibrationProjector(salobj.BaseScript):
+    """Move the calibration projector into a safe position
+
+    Parameters
+    ----------
+    index : int
+        Index of Script SAL component.
+    """
+
+    def __init__(self, index):
+        super().__init__(
+            index=index,
+            descr="Park Calibration Projector",
+        )
+
+        self.mtcalsys = None
+
+    def set_metadata(self, metadata):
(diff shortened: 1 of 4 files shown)


python/lsst/ts/externalscripts/maintel/park_calibration_projector.py
```diff
@@ -0,0 +1,139 @@
...
+        self.log.info("Configure completed")
+
+    async def run(self):
+        """Run script."""
+        await self.assert_components_enabled()
+
+        self.log.info("Parking Calibration Projector")
+        await self.mtcalsys.park_projector()
+
+        params = await self.mtcalsys.get_projector_setup()
```
edennihy, line 87: Is there something we could check/assert here after parking? I know we output the info below, but it might be nice to see the script fail if we don't fully park.

python/lsst/ts/externalscripts/maintel/park_calibration_projector.py
```diff
@@ -0,0 +1,139 @@
...
+            summary_state = await comp.evt_summaryState.aget()
+            try:
+                summaryState = summary_state.summaryState
+            except Exception as e:
+                self.log.debug(f"Exception: {e}")
+                summaryState = summary_state
+            if salobj.State(summaryState) != salobj.State(salobj.State.ENABLED):
+                raise RuntimeError(f"{comp} is not ENABLED")
+
+    @property
```
edennihy, line 121: If you use the assert_all_enabled, we won't need these properties.

python/lsst/ts/externalscripts/maintel/park_calibration_projector.py
```diff
@@ -0,0 +1,139 @@
...
+        if self.mtcalsys is None:
+            self.log.debug("Creating MTCalSys.")
+            self.mtcalsys = MTCalsys(domain=self.domain, log=self.log)
+            await self.mtcalsys.start_task
+
+        self.log.info("Configure completed")
+
+    async def run(self):
+        """Run script."""
+        await self.assert_components_enabled()
```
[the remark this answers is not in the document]
parfa30: Again, I cannot assume that the TunableLaser will be Enabled for this procedure.

python/lsst/ts/externalscripts/maintel/park_calibration_projector.py
```diff
@@ -0,0 +1,131 @@
...
+
+        await self.mtcalsys.get_projector_setup()
+
+        led_location = await self.mtcalsys.linearstage_led_select.tel_position.next(
+            flush=True
+        )
+        self.log.debug(f"LED Location: {led_location}")
+        assert led_location == self.mtcalsys.led_rest_position
+
+    async def assert_components_enabled(self):
```
edennihy, line 104: No need for a specific method here, you can just call self.mtcalsys.assert_all_enabled() directly.

tests/maintel/test_park_calibration_projector.py
```diff
@@ -0,0 +1,144 @@
...
+        await self.mock_mtcalsys()
+
+        self.log.debug("Finished initializing from basic_make_script")
+        return (self.script,)
+
+    async def mock_mtcalsys(self):
+        """Mock Calsys CSCs"""
+        self.script.mtcalsys = unittest.mock.AsyncMock()
+        self.script.mtcalsys.assert_liveliness = unittest.mock.AsyncMock()
+        self.script.mtcalsys.assert_all_enabled = unittest.mock.AsyncMock()
```
edennihy, line 61: Once you make the update to use this method, you should not need any of the .evt_summaryState.aget mocks.

parfa30:
Very clever using `params[1`]! Thanks for rehashing the tests. Made a small change, then squashed the commits.

edennihy (approved):

PR #186, opened 2025-04-24
Add parameter_march_triplet_lsstcam.py

gmegh:


edennihy (commented):
Thanks Guillem, the script itself looks pretty good. Be sure to also update the __init__.yaml file to import it. 

This script builds heavily on the parameter_march class, so there is good unit test coverage for most methods (e.g. https://github.com/lsst-ts/ts_externalscripts/blob/develop/tests/maintel/test_maintel_parameter_march_lsstcam.py). 

Still, could you please add some basic unit tests (the absolute minimum would be test_exectuable)?

edennihy (approved):
Approved!

PR #192, opened 2025-05-22
DM-51059: adding mtcs to lsstcam config so that filter will change

parfa30:


edennihy (approved):
Approved, but could you please add the script name that you are changing to the news fragment and commit message? So, instead of 
`adding mtcs to lsstcam config so that filter will change`

Something like
`In take_whitelight_flats_lsstcam.py, adding mtcs to lsstcam config so that the filter will change.`

PR #195, opened 2025-05-29
DM-51157: added salIndex to the groupId in base_make_calibrations

parfa30:


edennihy (approved):
Approved pending BTS test.

PR #202, opened 2025-06-23
DM-51498: Adding electrometer and fiberspectraphs as optional in the script

parfa30:


python/lsst/ts/externalscripts/maintel/take_calsys_flats_lsstcam.py
```diff
@@ -74,12 +76,14 @@ def get_schema(cls):
...
-                description: Will you use the camera during these flats
+                description: Will you use the camera during these flats.
+                             It overrides what is written in the mtcalsys.yaml
+                             configuration.
                 type: boolean
                 default: True
               random_seed:
                 description: Override of random seed for PTC Exposure list generation
-                type: [integer, "null"]
+                type: ["integer", "null"]
```
edennihy: I don't think quotes are required around type definitions that are not `"null"` (null is special in JSON). You can see that it is `type: [integer, "null"]` on the next line. 

I think both are correct, but can we drop this change to keep them consistent?

python/lsst/ts/externalscripts/maintel/take_calsys_flats_lsstcam.py
```diff
@@ -93,6 +97,22 @@ def get_schema(cls):
                 type: array
                 items:
                   type: string
+                default: []
```
edennihy: We don't typically set a default value for the `ignore:` parameter. You can remove this line.

tests/maintel/test_take_calsys_flats_lsstcam.py
```diff
@@ -300,6 +300,9 @@ async def test_configure_ignore(self):
                 {
                     "sequence_names": ["whitelight_r_57_dark"],
                     "use_camera": True,
+                    "use_electrometer": True,
```
edennihy, line 336: It would be great to add a separate unit test for these new configs (and for example check to make sure the fiberspectrograph cmds aren't sent when False), but adding them here doesn't really do much since this unit test only checks the components that were disabled by the ignore feature.

python/lsst/ts/externalscripts/maintel/take_calsys_flats_lsstcam.py
```diff
@@ -93,6 +97,22 @@ def get_schema(cls):
...
+              use_electrometer:
+                description: Will you use the electrometer in these tests.
```
edennihy: It might be worth adding to the description here that this also overrides the mtcalsys config and applies to _all_ electrometers, so if the mtcalsys config has 

  use_cbp_electrometer: True
  use_flatfield_electrometer: False

it will effectively make both of them False.

parfa30:
@edennihy I realized I also needed to update setup_calsys_flats.py. I'd appreciate another look at this when you have time!

PR #205, opened 2025-08-21
Tickets/osw 897 Ensure calibration scripts are enforcing required MTCS state

iglesu:
Add assert_feasibility hooks to ensure MTDomeTrajectory is ENABLED and MTDome is at least DISABLED before taking flat-field calibrations. This prevents vignetting state from being UNKNOWN and data landing on embargoed rack.
  

- Add assert_feasibility() to BaseTakeImage and BaseMakeCalibrations
- Implement dome state checks in TakeImageLSSTCam, MakeLSSTCamCalibrations, and TakeWhiteLightFlatsLSSTCam
- Unit tests

### DIFF
--- a/doc/news/OSW-897.feature.rst
+++ b/doc/news/OSW-897.feature.rst
@@ -1,0 +1,1 @@
+Ensure scripts used to take calibrations for SimonyiTel are enforcing the required MTCS state.


python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -198,3 +202,37 @@ def get_schema(cls):
...
+        """Ensure dome components are in the required state before flats."""
+        if image_type != "FLAT":
+            return None
+
+        # TODO: Review whether to add _dome_group initialization to
+        # start_remotes(). The start_remotes() method exists but isn't
+        # called anywhere in this script.
+
+        if self._dome_group is None:
+            self._dome_group = RemoteGroup(
```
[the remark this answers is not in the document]
iglesu: I believe there isn't a MTCS in the `make_lsstcam_calibrations.py`. Should I instantiate one such as in `[take_whitelight_flats_lsstcam.py](https://github.com/lsst-ts/ts_externalscripts/blob/48af07b18e5720e7d26aabede0547168748364f4/python/lsst/ts/externalscripts/maintel/take_twilight_flats_lsstcam.py#L53-L59)`?  Would that be more efficient?

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -47,6 +51,7 @@ def __init__(self, index=1):
         )
         self._lsstcam = None
         self._ocps_group = None
+        self._mtcs = None
```
tribeiro, line 54: So, the reason we have `lsstcam` and `ocps_group` being "private" (the leading "\_") is because the `camera`  and `ocps_group` are properties for the abstract class. For the `MTCS` I we don't have such constraint so I think you can drop the "\_".

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -104,6 +109,27 @@ async def start_remotes(self):
             )
             await self._ocps_group.start_task
 
+        if not hasattr(self, "_mtcs") or self._mtcs is None:
+            self._mtcs = MTCS(domain=self.domain, log=self.log)
```
tribeiro, line 113: So, the way `MTCS` is defined here it will subscribe to all the topics from all the components. This is usually pretty heavy. To limit resources you can provide a `intended_usage`, something like:

```
self._mtcs = MTCS(
    domain=self.domain,
    log=self.log,
    intended_usage=MTCSUsages.Slew,
)
```

you will have to import `MTCSUsages`,

```
from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
```

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -182,6 +208,16 @@ def get_schema(cls):
...
                 default: "/repo/LSSTCam/butler+sasquatch.yaml"
+            ignore:
+                description: >-
+                  CSCs from the MTCS group to ignore in status check. Name must
+                  match those in self._mtcs.components_attr, e.g.; mtmount, mtptg.
+                  By default, all components except mtdometrajectory and mtdome are ignored.
+                type: array
+                items:
+                  type: string
+                default: ["mtmount", "mtptg", "mtaos", "mtm1m3", "mtm2", "mthexapod_1",
```
tribeiro, line 219: Please, remove default.

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -198,3 +234,45 @@ def get_schema(cls):
...
         return dict(filter=self.config.filter)
+
+    async def assert_feasibility(self, image_type):
+        """Ensure dome components are in the required state before flats."""
+        if image_type != "FLAT":
+            return None
+
+        # Check if dome components are ignored by user configuration.
+        if (
+            hasattr(self._mtcs.check, "mtdometrajectory")
```
tribeiro, line 245: you don't need the `hasattr(self._mtcs.check, "mtdometrajectory")`

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -198,3 +234,45 @@ def get_schema(cls):
...
+            and not self._mtcs.check.mtdometrajectory
+        ):
+            raise RuntimeError(
```
tribeiro, line 248: We should allow people to ignore mtdometrajectory, please remove this exception.

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -198,3 +234,45 @@ def get_schema(cls):
...
+                "Cannot take FLAT images when MTDomeTrajectory is ignored. "
+                "Dome trajectory state is required to ensure proper vignetting state publication."
+            )
+
+        if hasattr(self._mtcs.check, "mtdome") and not self._mtcs.check.mtdome:
+            raise RuntimeError(
```
tribeiro, line 254: We should allow people to ignore the dome, please, remove this exception.

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -198,3 +234,45 @@ def get_schema(cls):
...
+            self._mtcs.rem.mtdome.evt_summaryState.aget(
+                timeout=self._mtcs.long_timeout
+            ),
+        )
+        dome_trajectory_summary_state = salobj.State(dome_trajectory_evt.summaryState)
+        dome_summary_state = salobj.State(dome_evt.summaryState)
+
+        if dome_trajectory_summary_state != salobj.State.ENABLED:
+            raise RuntimeError(
+                "MTDomeTrajectory must be ENABLED before taking FLATs to ensure "
```
tribeiro, line 272: You should add the current state:

```
raise RuntimeError(
    "MTDomeTrajectory must be ENABLED before taking FLATs to ensure "
    "vignetting state is published. "
    f"Current state {dome_trajectory_summary_state.name}."
)
```

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -198,3 +234,45 @@ def get_schema(cls):
...
+                "vignetting state is published."
+            )
+        if dome_summary_state not in (salobj.State.DISABLED, salobj.State.ENABLED):
```
tribeiro, line 275: Instead of using a `tuple` use a `set`... Comparison on a `set` is more performant:

```
if dome_summary_state not in {salobj.State.DISABLED, salobj.State.ENABLED}:
```

python/lsst/ts/externalscripts/maintel/make_lsstcam_calibrations.py
```diff
@@ -198,3 +234,45 @@ def get_schema(cls):
...
+            raise RuntimeError(
+                "MTDome must be at least DISABLED before taking FLATs (not OFFLINE/STANDBY/FAULT)."
```
tribeiro, line 277: Consider putting the acceptable states in a variable then doing:

```
acceptable_mtdome_state = {salobj.State.DISABLED, salobj.State.ENABLED}

if dome_summary_state not in acceptable_mtdome_state:

    raise RuntimeError(
                f"MTDome must be in {acceptable_mtdome_state} before taking flats, current state {dome_summary_state.name}."
    )
```

tribeiro (approved):

PR #231, opened 2026-03-25
Fix the warmup_hexapod.py that all movements should be sync.

teweitsai:


dsanmartim (approved):
It looks good to me. Tks!

PR #235, opened 2026-05-26
tickets/RSO-559: Add TCBP scan script

elanaku:


parfa30 (approved):
Approved based on the fact that this has been run successfully on the summit.

python/lsst/ts/externalscripts/maintel/tcbp_scan_lsstcam.py
```diff
@@ -0,0 +1,96 @@
...
+        return schema_dict
+
+    def get_image_id_key(self):
+        return "LSSTCAMID"
+
+    @staticmethod
+    def _normalize(v):
+        return str(v).lower() if v is not None else "empty"
+
+    def default_filter_for_optimized_wls(self, config):
```
b1quint: Small ask - does `wls` mean wavelengths? If so, could you rename it so it is easier to read? It is a small thing. But when I read the code, I have to pause, look around, and confirm if `wls` really means what I think it means. Optional.

python/lsst/ts/externalscripts/maintel/tcbp_scan_lsstcam.py
```diff
@@ -0,0 +1,96 @@
...
+        """Map the LSSTCam filter onto an optimized-wavelength key.
+
+        LSSTCam filters use names like ``r_57`` that don't match the
+        :func:`get_optimized_wls` keys; fall back to the broad ``"empty"``
```
b1quint: Same with `get_optimized_wls`. If the change becomes too big, it might be worth pushing for later. Again, this is just to improve readability. Light suggestion.

PR #241, opened 2026-06-30
add self.mtcs to LSSTCam

parfa30:


parfa30:
tested on BTS

elanaku (approved):
Looks good.

